PR #18, opened 2020-12-01
Update Project Structure

pkrawc:
Hey @quaardvark,

I'm opening this PR as a bit of a request for comment, I still have trouble stepping through the composition of some of the slices/components in the app now that we've integrated Prismic.

I don't know if having them live in two spots (two components named the same but in different locations) is really ideal. I get the want for separation of concerns but I think that it might be slightly over-optimizing and also I'm not convinced any of the `core` components can be used for anything but slices, wanting to keep them as far away as possible from the slice implementation isn't beneficial in my mind.

Anyway, I've updated some of the _app and _document files to better consume google tags and moved the default SEO into a config file.

I've also created a `src` folder, in this I've laid out what I think is a good atomic folder structure with some blank files that try and illustrate what type of components should go where.

Let me know what downsides to this you can see.


quaardvark:
There's a lot of idiosyncratic styling in Slices, which is more specific to recreation of the current LPs than to what these pages would look like when designed properly. I figured that separate Slice components that implement generic elements would be a decent place to dump this specificity. 

Speaking in atomic design terms, the slice is a higher order organism than core component since it adds has 2 specific couplings - Prismic API and project/page type specific styling.  I can see the core components being re-used in different slices, - for newly designed pages, while keeping older slices in tact. Especially useful in the context of testing against control.

Maybe we should just rename slice components to reflect their specificity, but still keep them separated?

pkrawc:
Are you saying that we'll eventually not need the prismic slice components?

I agree that the shape of the prismic data is fairly specific but probably shouldn't need another component, I bet the implementation details or reshaping of the data can happen in what is essentially the large switch statement for a slice zone.

I also think there's something to just building the organism (slice) components to use the prismic data shape anyway and require any non-prismic implementation of the organism to be shaped the same way.

quaardvark:
I'm saying the opposite, - eventually there would be more Slice components than there are Core components. For example a Hero Unit, - in a new Prismic repo I created a slice called "Hero w/ Email capture" that combines Hero and Email Capture core components in one slice. There's another slice called "Email Capture" that has just a Email Capture core component paired with title and hash anchor.

While typing this I realized that I'm actually completely against merging Slices and Core components, - this will inevitably lead to monstrous components that support every possible combination of Prismic/CMS specific implementations.

quaardvark:
@pkrawc 
I made a go on restructuring, somewhat derived your Atomic Design suggestion, with some trade-offs concerning Prismic Slices and specific components. My suggestion is to add a notion of `partials` next to `organisms` and `molecules` as a variation on Atomic Design `templates`, - this folder would contain something that has more specificity than `organisms` and generally would directly compose a page, - like SliceZone, AvailFooter or PreviewWarning.

The reasoning is the following:
- Specificity - these components would generally not be the elements of the design library, but a specific concretions required for specific pages, or work with CMS-specific DSL. For example: we can expect at some point to come up with generic `Footer` `organism`, which would support dynamic content, while `AvailFooter` is a very specific and static `partial`
- Variations - `organisms` and below would likely to have some design variations and Storybooks associated with them, while `partials` being very specific, - would not
- Separation of CMS-specific requirements from design library. If content creators would like their specific Slice to have 15rem padding around it, - we should be able to slap it without worrying about affecting other Slices / variations

With that reasoning, I've put Slices and Prismic-specific helpers, such as RichText customization as a subcomponents directory under SliceZone, - I'm not a fan of how deep they are, but at the moment it makes the most sense to me. To keep thing clearer, I think we should have a separate naming convention for Slices, that will convey exact concretion of what it is in CMS, something like `ThingItsCalledInCMSSlice`.

I'm still strongly convinced in keeping Slices as specific concretions of design library `organisms`, at least for now, because of aforementioned reasoning and scaling concerns. Think of Slices as: 
1) binding between CMS DSL and somewhat abstract design elements/organisms, - this not only concerns the API data model on its own, but also things like RichText; 
2) a space to dump specific asks to change styling for specific CMS Slices, - we can wrap / inject styling into the `organism` without affecting it at the core;
3) a space to implement unusual / specific things, - look at the `PitchCardsSlice` with the `titleImage` - for the TransUnion logo in the title, - I think implementing it in the `organism` of `PitchCards` would be leaky. A better way would have been to create another `organism` or `molecule`, - a heading with inlined image and compose these two `organisms` into a Slice;
4) a space to compose / re-use `organisms` in different, CMS/content-specific ways, like the `HeroWithEmailCaptureSlice`;

PR #27, opened 2020-12-15
Creates Hero Slice with buttons

henriquecfreitas:


src/components/elements/Link/index.js
```diff
@@ -0,0 +1,25 @@
...
+  }, []);
+
+  function isInternal(link) {
+    const escapedHost = host.replace('www.', '').replace(/\./, "\\.")
+    const internalLinkRegex = new RegExp(`^(http(s?)://)?(www\\.)?(${escapedHost})?(/(.)*)?$`)
+    return internalLinkRegex.test(link)
+  }
+
+  return isInternal(href) ?
+    <NextLink href={href}>
```
quaardvark, line 17: Next Link supports external URLs since v10, so this split is unnecessary

quaardvark (changes requested):

src/components/partials/SliceZone/components/HeroSlice/index.js
```diff
@@ -0,0 +1,59 @@
...
+          <Link link={primaryButtonLink}>
+            <Button background={background} isPrimary>
+              {primaryButtonText}
+            </Button>
+          </Link>
+        )
+      }
+      secondaryLink={
+        showSecondaryButton && (
+          <Link link={secondaryButtonLink}>
```
quaardvark, line 50: `<a><button /></a>` is not a valid HTML, to fix this we should:

Refactor `Link` as such:
```
<NextLink passHref href={href} {...props}>
  {React.cloneElement(children, {
    ...(target && { target, rel: "noopener" }),
  })}
</NextLink>
```

Use it in the slice as:
```
<Link link={primaryButtonLink}>
  <Button forwardedAs="a" background={background} variant="primary">
    {primaryButtonText}
  </Button>
</Link>
```

src/components/partials/SliceZone/components/Button/index.js
```diff
@@ -0,0 +1,32 @@
...
+
+function Button({ background, isPrimary = false, children, ...props }) {
+  const [scheme, isDark] = background ? analyzeColor(background) : []
+
+  let buttonElement = AvailButton
+  if (isDark) {
+    buttonElement = isPrimary ? ContrastButtonPrimary : ContrastButton
+  }
+
+  return React.createElement(
```
quaardvark, line 18: We should use `cloneElement` instead of `createElement` for performance reasons. You might need to change references to elements to JSX

src/components/partials/SliceZone/components/Link/index.js
```diff
@@ -0,0 +1,46 @@
...
+  Document: "Document",
+  Media: "Media",
+}
+
+const Link = ({ link, children, ...props }) => {
+  const linkType = link.link_type
+
+  const urlResolver = useUrlResolver()
+
+  const { href, target } = (() => {
```
quaardvark, line 16: Great job here!

src/components/partials/SliceZone/components/Button/index.js
```diff
@@ -0,0 +1,32 @@
...
+    buttonElement,
+    {
+      scheme,
+      variant: isPrimary ? "primary" : "default",
```
quaardvark, line 22: We should just forward `variant` instead of using `isPrimary`

quaardvark (changes requested):

src/components/partials/SliceZone/components/Button/index.js
```diff
@@ -0,0 +1,32 @@
...
+      display: "block",
+      textAlign: "center",
+      width: "100%",
+      ...props,
+    },
+    children
+  )
+}
+
+export default Button
```
quaardvark, line 32: We should use `React.forwardRef` here is pass the ref to `cloneElement`

src/components/partials/SliceZone/components/Button/index.js
```diff
@@ -0,0 +1,32 @@
+import React from "react"
```
quaardvark, line 1: I think we can move this button to `elements` and merge `ContrastButton` into it, since it doesn't have use CMS specific API

quaardvark (approved):

PR #31, opened 2020-12-18
Remove Prismic Transpile in Config

pkrawc:
Hey @quaardvark,

I've updated storybook and some outdated dependencies (and visually tested in storybook) as well as removed the next.config.js file, looking at the node_modules folder for prismic-reactjs we shouldn't need nextjs to compile it I don't think.


pkrawc:
I love it, by the time I was done writing the PR Zeit had confirmed that it would build the app without a config just fine.

quaardvark:
@pkrawc I don't remember why transpiling was there it the first place, so I'll QA it. Thanks for taking the time to upgrade the deps :+1:

quaardvark (approved):

PR #33, opened 2020-12-21
Refactor NavBar to simplify the input required and to facilitate "automatic" responsiveness

quaardvark:
This PR simplifies NavBar in the following ways:

- Removes `breakpoint` options
- Facilitates "automatic" responsiveness

Introduces the following limitations:

- Only one primary button can exist - the first button with `primary: true`
- Priority of buttons for responsiveness is simply controlled by their order and behaves differently when there are "pushed" buttons present
    - With no pushed buttons, - the order is conventional - the first button will be hidden last
    - With pushed buttons present, - the order is reversed - the last (pushed) buttons will be hidden last

"Automatic" responsiveness is based on `flex-wrap` and `overflow: hidden`, where buttons wrap out of view when they no longer fit in the box.

### DIFF
--- a/pages/info/[uid].js
+++ b/pages/info/[uid].js
@@ -92,12 +92,15 @@ const Page = ({ data, uid }) => {
 
   const navBarLinks =
     data.nav_bar &&
-    data.nav_bar.map(({ buttonText, buttonLink, buttonHash, primary }) => ({
-      text: buttonText,
-      href: buttonHash ? `#${buttonHash.replace(/^#/, "")}` : buttonLink.url,
-      target: buttonLink && buttonLink.target,
-      primary,
-    }))
+    data.nav_bar.map(
+      ({ buttonText, buttonLink, buttonHash, primary, push }) => ({
+        text: buttonText,
+        href: buttonHash ? `#${buttonHash.replace(/^#/, "")}` : buttonLink.url,
+        target: buttonLink && buttonLink.target,
+        primary,
+        push,
+      })
+    )
   return (
     <UrlResolverProvider params={urlResolverParams}>
       <NextSeo
--- a/pages/info/[uid].js
+++ b/pages/info/[uid].js
@@ -92,15 +92,12 @@ const Page = ({ data, uid }) => {
 
   const navBarLinks =
     data.nav_bar &&
-    data.nav_bar.map(
-      ({ buttonText, buttonLink, buttonHash, primary, push, breakpoint }) => ({
-        text: buttonText,
-        href: buttonHash ? `#${buttonHash.replace(/^#/, "")}` : buttonLink.url,
-        primary,
-        push,
-        breakpoint,
-      })
-    )
+    data.nav_bar.map(({ buttonText, buttonLink, buttonHash, primary }) => ({
+      text: buttonText,
+      href: buttonHash ? `#${buttonHash.replace(/^#/, "")}` : buttonLink.url,
+      target: buttonLink && buttonLink.target,
+      primary,
+    }))
   return (
     <UrlResolverProvider params={urlResolverParams}>
       <NextSeo
--- a/src/components/organisms/NavBar/index.js
+++ b/src/components/organisms/NavBar/index.js
@@ -2,6 +2,7 @@ import React, { useMemo } from "react"
 import { Box, Container, Flex } from "@rent_avail/layout"
 import { Button } from "@rent_avail/controls"
 import styled, { createGlobalStyle } from "styled-components"
+import { getTargetProps } from "utils/link"
 
 const GlobalStyle = createGlobalStyle`
   :root {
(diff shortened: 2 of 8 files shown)


src/components/organisms/NavBar/index.js
```diff
@@ -67,33 +64,51 @@ const NavBar = ({
...
-                : "block"
-            return (
-              <React.Fragment key={link.href}>
-                {index === splitIndex && <Box ml="auto" key="separator" />}
+          <Flex
+            sx={{
+              flexDirection: "row",
+              justifyContent: primaryLink
+                ? "flex-start"
+                : ["flex-end", "flex-end", "flex-end", "flex-start"],
```
henriquecfreitas, line 72: I would suggest to remove those ternaries from the props and to place it before the render function, in order to make the code more clear. If you place it in that `links` `useMemo` callback it would be even better, because the ternary wouldn't be performed on each render, right?

```
const defaultFlexJustifyContent = [ "flex-end", "flex-end", "flex-end", "flex-start" ]
const primaryNavBarButtonDisplay = [ "none", "none", "block" ]
const [ flexJustifyContent, navBarButtonDisplay ] = primaryLink
                  ? [ "flex-start",  primaryNavBarButtonDisplay ]
                  : [ defaultFlexJustifyContent, "block" ]
```
```suggestion
              justifyContent: {flexJustifyContent},
```

src/components/organisms/NavBar/index.js
```diff
@@ -67,33 +64,51 @@ const NavBar = ({
...
-                  variant={link.primary ? "primary" : "default"}
-                  href={link.href}
+                  key={href}
+                  href={href}
+                  {...(target && { target, rel: "noopener" })}
                   forwardedAs="a"
-                  display={display}
-                  mr={index === links.length - 1 ? "0" : "2rem"}
+                  ml="2rem"
+                  display={primaryLink ? ["none", "none", "block"] : "block"}
```
henriquecfreitas, line 87:
```suggestion
                  display={navBarButtonDisplay}
```

henriquecfreitas (approved):

src/components/organisms/NavBar/index.js
quaardvark: I'm not sure if that makes the code clearer. If anything it introduces the gap between declaration and application, which doesn't work at all for CSS. The only real reason to alias CSS properties via variables is for DRY, which is not the case here.

Ternaries are VERY fast, there's no real reason to optimize them.

pages/info/[uid].js
```diff
@@ -92,15 +92,12 @@ const Page = ({ data, uid }) => {
...
-        href: buttonHash ? `#${buttonHash.replace(/^#/, "")}` : buttonLink.url,
-        primary,
-        push,
-        breakpoint,
-      })
-    )
+    data.nav_bar.map(({ buttonText, buttonLink, buttonHash, primary }) => ({
+      text: buttonText,
+      href: buttonHash ? `#${buttonHash.replace(/^#/, "")}` : buttonLink.url,
+      target: buttonLink && buttonLink.target,
```
Comment (pkrawc) [inline pages/info/[uid].js:98]: Is undefined an accepted return for a target? I would think that this should be the following.
```
buttonHash?.replace(/^#/, "") || _self
```

src/components/organisms/NavBar/index.js
```diff
@@ -14,23 +14,20 @@ const NavBarButton = styled(Button)`
   height: 48px;
 `
 
-const NavBar = ({
+export default function NavBar({
   background = "ui_100",
   links = [],
```
pkrawc, line 19: Reading through this file I generally think the shape of a link is fairly opaque. It took me a while to grok that there is going to be a `primary` key on one (and only one?) link and we're sorting based on that.

I think just documenting the expected API for links would be good here, so the user doesn't need to understand intrinsically what the shape of the link's Prismic data is.

src/components/organisms/NavBar/index.js
```diff
@@ -49,12 +46,12 @@ const NavBar = ({
...
             as="a"
             href="https://avail.co"
-            mr="2rem"
-            width={["48px", "auto"]}
-            height="48px"
             sx={{
               display: "block",
               overflow: "hidden",
+              width: ["48px", "auto"],
+              height: "48px",
```
pkrawc, line 53: Should probably use the space theme key or `rem`s to make sure this component stays themeable.

src/components/organisms/NavBar/index.js
```diff
@@ -67,33 +64,51 @@ const NavBar = ({
...
-                {index === splitIndex && <Box ml="auto" key="separator" />}
+          <Flex
+            sx={{
+              flexDirection: "row",
+              justifyContent: primaryLink
+                ? "flex-start"
+                : ["flex-end", "flex-end", "flex-end", "flex-start"],
+              flex: "1 1 auto",
+              flexWrap: "wrap",
+              height: "48px",
```
pkrawc, line 75: Again on the pixel values. If pixels are needed we should be utilizing the space key.

src/components/organisms/NavBar/index.js
```diff
@@ -67,33 +64,51 @@ const NavBar = ({
...
+            }}
+          >
+            {defaultLinks.map(({ href, text, target }) => {
+              return (
                 <NavBarButton
-                  variant={link.primary ? "primary" : "default"}
-                  href={link.href}
+                  key={href}
+                  href={href}
+                  {...(target && { target, rel: "noopener" })}
```
pkrawc, line 84: Ah, so we're checking for the target here as well. You might be able to remove the duplicate check by just coercing the link props earlier. Maybe this is overkill though, your call.

pkrawc (changes requested):
I think the overflow hidden thing and a flex-wrap container is really clever. I probably would've opted for calculating the available width with js but your solution seems much more elegant.

We lose the functionality of being able to push a nav item to the right, yeah? The pricing link in the below diagram.
```
[logo] [thing one] [thing two] ------------- [pricing] [primary cta]
```

src/components/organisms/NavBar/index.js
quaardvark: Added some comments to document that

pages/info/[uid].js
Comment (quaardvark) [reply pages/info/[uid].js]: Yeah, it should simply not set `target` and `rel=noopener` and if it's undefined. Can be a jump-link hash

src/components/organisms/NavBar/index.js
quaardvark: Done

src/components/organisms/NavBar/index.js
quaardvark: Done

src/components/organisms/NavBar/index.js
quaardvark: I moved it into `utils`, since there are few components that use the exact same pattern, so the API is now `{...getTargetProps(target)}` and `getTargetProps` omits `noopener` for `_self` links just in case. I don't really like to mess with `target` and related attributes, since it may have unintended consequences in different browsers, so I prefer to not set anything if `target` is undefined

PR #34, opened 2020-12-23
Creating slider component and testimonials slice

henriquecfreitas:


src/components/molecules/Slider/index.js
```diff
@@ -0,0 +1,131 @@
...
+    if (currentIdx < refs.length - 1) scrollTo(currentIdx + 1)
+  }
+
+  return (
+    <React.Fragment>
+      <SliderBox ref={slideRef} onScroll={calcVisibilityIndexes}>
+        <SliderStack wrapChildren direction={["row"]}>
+          {children.map((childElement, idx) => {
+            refs[idx] = createRef()
+            return cloneElement(childElement, { ref: refs[idx] })
```
quaardvark, line 105: I get a lot of warnings on missing `key` prop

quaardvark (changes requested):

src/components/molecules/Slider/index.js
```diff
@@ -0,0 +1,131 @@
...
+
+function Slider({ children }) {
+  const slideRef = createRef()
+  const refs = []
+
+  const [currentIdx, setCurrentIdx] = useState(null)
+  const [firstVisibleIdx, setFirstVisibleIdx] = useState(null)
+  const [lastVisibleIdx, setLastVisibleIdx] = useState(null)
+
+  function calcVisibilityIndexes() {
```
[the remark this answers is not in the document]
henriquecfreitas: The scroll function were modified to scroll to the middle of the next/last element that is not fully visible

src/components/molecules/Slider/index.js
```diff
@@ -0,0 +1,131 @@
...
+      <SliderBox ref={slideRef} onScroll={calcVisibilityIndexes}>
+        <SliderStack wrapChildren direction={["row"]}>
+          {children.map((childElement, idx) => {
+            refs[idx] = createRef()
+            return cloneElement(childElement, { ref: refs[idx] })
+          })}
+        </SliderStack>
+      </SliderBox>
+      <ScrollControlsContainer>
+        <ChevronLeft
```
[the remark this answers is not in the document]
henriquecfreitas: It is disappearing now, keeping the same height for the controls box that it would have if the buttons were visible

pkrawc:
Hey guys, sorry I've silent on this. I'm planning on digging tomorrow morning at some point but generally this isn't how I would've gone about creating a carousel. I think we can likely leverage the IntersectionObserver api to get a list of visible items without the need for manual calculation. I think it's totally fair to corner Nellie and lean on her a bit more to illustrate the interactions expected if we're running into a wall on that.

There's no reason to use things like `styled(Box)` since box really doesn't have any default styling and can take any style key in the `sx` prop.

The `Carousel` component is a really good candidate to split out into a hook that controls ui logic and a component that controls style.

```jsx
function useCarousel(props) {
  // return a list of items that should be shown
  // useEffect to check whatever
  return {
    // Carousel Props
  }
}

function Carousel(props, ref) {
  const htmlProps = useCarousel({...props, ref})
  return <Box className="carousel" sx={{/* Carousel Styles */}} />
}
```

src/components/organisms/Testimonials/index.js
```diff
@@ -0,0 +1,90 @@
...
+  display: flex;
+  flex-flow: column;
+  align-items: center;
+  justify-content: center;
+  text-align: center;
+  border-radius: 12px;
+`
+
+function Testimonials({
+  bg,
```
pkrawc, line 22: Most props should have default values if possible, it's a good way to prevent render errors.

quaardvark:
@pkrawc are you suggesting to run an IntersectionObserver for each child to detect which are clipping? Would it really be cleaner than calculations?

pkrawc:
@quaardvark I'm not actually sure. I have a pairing session with @henriquecfreitas this morning to sync up.

I dug more into this slice last night and calculations might be the way to go.

I also wonder if there's more than just scroll position we want to change of the selected testimonial. Which is as much a design thing as an engineering thing.

henriquecfreitas:
> @pkrawc are you suggesting to run an IntersectionObserver for each child to detect which are clipping? Would it really be cleaner than calculations?

I think is way better now, handling the visibility of the items with a callback for the IntersectionObserver

```
  const [ visibilityIndexes, setVisibilityIndexes ] = useState([])
  const firstVisibleIdx = visibilityIndexes.indexOf(true)
  const lastVisibleIdx = visibilityIndexes.lastIndexOf(true)

  function observerCb(entries) {
    entries.forEach(entry => {
      const index = refs.current.indexOf(entry.target)
      const isVisible = entry.intersectionRatio > VISIBLE_MIN_INTERSECTION
      const updatedVisibilityIndexes = visibilityIndexes
      updatedVisibilityIndexes[index] = isVisible
      setVisibilityIndexes([...updatedVisibilityIndexes])
    })
  }
  useEffect(() => {
    const observer = new IntersectionObserver(observerCb, {
      threshold: VISIBLE_MIN_INTERSECTION,
    })
    refs.current.forEach(ref => observer.observe(ref))
  }, [])
```

pkrawc:
@henriquecfreitas yeah that's exactly what I was imagining. We should probably set the observer as a ref instead of creating one on ever useEffect call but this is the main gist.

henriquecfreitas:
> I also wonder if there's more than just scroll position we want to change of the selected testimonial. Which is as much a design thing as an engineering thing.

Maybe applying a scale, like that?
Screen Shot 2020-12-30 at 10 47 10

quaardvark:
@henriquecfreitas IIRC entries will be given only for updated values when observer callback fires, so you'll need a bit more complex logic to track visibility

quaardvark:
@pkrawc @henriquecfreitas I'm still not sure on the UX here, - don't we want to track intersection with `Container` and not just `window`? Assuming that the arrows should scroll until either left or right extremes are within the `Container`.

pkrawc:
Good catch, yeah probably the container. You can set the root element in the options when creating the observer

henriquecfreitas:
> @pkrawc @henriquecfreitas I'm still not sure on the UX here, - don't we want to track intersection with `Container` and not just `window`? Assuming that the arrows should scroll until either left or right extremes are within the `Container`.

No, I think we need to track the intersection with the Carousel extended view (including the side scroll spaces), not the Container. The windows is working on this case because the carousel 's horizontal view is exactly the same as the window's horizontal view.
 But you're right, it can't be based on the window. I'll set the carousel box as the root view and use the `scrollSpace` prop values as the rootMargin

henriquecfreitas:
> IIRC entries will be given only for updated values when observer callback fires, so you'll need a bit more complex logic to track visibility

It also fires the callback after the setting the observer for the element, so I can get its initial visibility

quaardvark:
@pkrawc @henriquecfreitas I can't help but think that these components, calculations, observers are unnecessary complex. I tried to come up with the most simple solution I could think of atm - https://codesandbox.io/s/carousel-v3-kiss-8q3x2?file=/src/index.js

It basically use uses two very simple functions for scrolling:
```
const scrollLeft = () => {
  const leftClipped = childrenRef.current
    .slice()
    .reverse()
    .find(({ offsetLeft, offsetWidth }) => {
      return offsetLeft - scrollRef.current.scrollLeft < 0;
    });
  if (leftClipped) {
    scrollRef.current.scrollTo(leftClipped.offsetLeft, 0);
  }
};

const scrollRight = () => {
  const rightClipped = childrenRef.current.find(
    ({ offsetLeft, offsetWidth }) => {
      return (
        offsetLeft + offsetWidth - scrollRef.current.scrollLeft >
        containerRect.width
      );
    }
  );
  if (rightClipped) {
    scrollRef.current.scrollTo(rightClipped.offsetLeft, 0);
  }
};
```

Not a big fan of left arrow UX, but I think we can tweak it

pkrawc:
> @pkrawc @henriquecfreitas I can't help but think that these components, calculations, observers are unnecessary complex. I tried to come up with the most simple solution I could think of atm - https://codesandbox.io/s/carousel-v3-kiss-8q3x2?file=/src/index.js
> 
> It basically use uses two very simple functions for scrolling:
> 
> ```
> const scrollLeft = () => {
>   const leftClipped = childrenRef.current
>     .slice()
>     .reverse()
>     .find(({ offsetLeft, offsetWidth }) => {
>       return offsetLeft - scrollRef.current.scrollLeft < 0;
>     });
>   if (leftClipped) {
>     scrollRef.current.scrollTo(leftClipped.offsetLeft, 0);
>   }
> };
> 
> const scrollRight = () => {
>   const rightClipped = childrenRef.current.find(
>     ({ offsetLeft, offsetWidth }) => {
>       return (
>         offsetLeft + offsetWidth - scrollRef.current.scrollLeft >
>         containerRect.width
>       );
>     }
>   );
>   if (rightClipped) {
>     scrollRef.current.scrollTo(rightClipped.offsetLeft, 0);
>   }
> };
> ```
> 
> Not a big fan of left arrow UX, but I think we can tweak it

This is good, observer was a simplification of the manual calculations but I'm liking this more. It's really calling into question the UX of the whole thing for me though.

I don't think the carousel component will be useless but I'm not actually sure if this is the interaction design we should go for. I can't help but think that users are just going to ignore all this text as superfluous, when if we showed them one testimonial at a time it would be more impactful.

src/components/organisms/Testimonials/index.js
```diff
@@ -0,0 +1,171 @@
...
+  text-align: right;
+  > * {
+    margin: 1rem;
+    opacity: 0.9;
+    &.scrollControlEnabled:hover {
+      cursor: pointer;
+      opacity: 1;
+    }
+    &.scrollControlDisabled {
+      opacity: 0.7;
```
quaardvark, line 33: Probably should go as low as 0.2-0.3 for visibility

src/components/organisms/Testimonials/index.js
```diff
@@ -0,0 +1,171 @@
...
+    setScrollSpace(parseInt(marginLeft, 10) + parseInt(paddingLeft, 10))
+  }, [containerRect])
+
+  useEffect(() => {
+    if (containerObserver) containerObserver.disconnect()
+
+    const cb = containerObserverCb(setLeftClipped, setRightClipped)
+    const options = {
+      root: containerRef.current,
+      rootMargin: `0px ${scrollSpace}px`,
```
quaardvark, line 95: `rootMargin` should be removed here. It makes it the same as having no root at all, - what's the point then? The idea was to track intersection with `Container` edges. Otherwise this can happen:

Screenshot_20210104_131045

Here testimonials are "scrolled" but both buttons are disabled, because there's no intersection with window edges

quaardvark (changes requested):

src/components/partials/SliceZone/components/TestimonialsSlice/index.js
```diff
@@ -0,0 +1,39 @@
...
+    ({ picture, author, titleAndLocation, quote }) => ({
+      picture,
+      author,
+      titleAndLocation,
+      quote: <RichText render={quote} />,
+    })
+  )
+
+  return (
+    <Testimonials
```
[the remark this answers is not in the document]
quaardvark: Did we address this? Is skew a CMS option or is it forced skew?

.eslintrc
```diff
@@ -16,6 +16,8 @@
     "semi": [1, "never"],
     "comma-dangle": [1, "only-multiline"],
     "consistent-return": [0],
+    "no-plusplus": ["error", { "allowForLoopAfterthoughts": true }],
+    "prefer-arrow-callback": [ "error", { "allowNamedFunctions": true } ],
```
quaardvark, line 20: If we don't use these, let's clean it up. In general, we prefer to override ESLint contextually in the file with `//eslint-disable rule` comments rather that changing the rules. There's no good reason for either of these changes

henriquecfreitas: I'll remove the `no-plusplus` rule right now.
@pkrawc can you give us a position on the `prefer-arrow-callback` rule?

src/components/organisms/Testimonials/index.js
henriquecfreitas: The buttons are disabled because all the content is visible, it will be enabled when you scroll enough to hide part of the first or the last card. We are tracking intersection with the `Container`, but it has margins on the left and right that need to be passed to the Intersection Observer constructor. Otherwise it would consider only the inner space of the Container, ignoring its margins as visible space

quaardvark: It should consider only the inner space, - that was the idea

.eslintrc
pkrawc: We need to leave the `allowNamedFunctions` boolean as true so that one can write
```jsx
const Comp = forwardRef(function Comp(props, ref) {})
```
If you don't use a function declaration for the inner function then in devtools and stack traces you'll see a lot of `forwardRef: function` instead of `forwardRef: Comp`

src/components/organisms/Testimonials/index.js
```diff
@@ -0,0 +1,171 @@
...
+}) {
+  const containerRef = useRef()
+
+  const [scrollSpace, setScrollSpace] = useState(0)
+  const containerRect = useResize(containerRef)
+
+  const [leftClipped, setLeftClipped] = useState(false)
+  const [rightClipped, setRightClipped] = useState(false)
+
+  const [containerObserver, setContainerObserver] = useState()
```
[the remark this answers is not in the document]
pkrawc: Agreed here. A couple tutorials have the observer set as state but that's really only important if you want to be notified of changes to the observer itself.

src/components/partials/SliceZone/components/TestimonialsSlice/index.js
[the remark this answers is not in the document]
henriquecfreitas: On the source it is addressed, but I still need to add the CMS option on prismic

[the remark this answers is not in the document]
henriquecfreitas: I just added it on the CMS, but it is not working, I'm into it right now

src/components/organisms/Testimonials/index.js
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  useEffect(() => {
+    const container = containerRef.current
+    if (!container || !getComputedStyle) return
+    const { marginLeft, paddingLeft } = getComputedStyle(container)
+    setScrollSpace(parseInt(marginLeft, 10) + parseInt(paddingLeft, 10))
+  }, [containerRect])
+
+  useEffect(() => {
```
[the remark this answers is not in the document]
henriquecfreitas: I didn't understood this one, can you explain it better please?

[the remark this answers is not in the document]
quaardvark: A function returned from `useEffect` is called when the component is being unmounted. It's intended use is to clean up any side effects set up within `useEffect`. We just need to return function that will disconnect the observer, so that it will be recycled properly. Otherwise it can potentially be a memory leak

src/components/organisms/Testimonials/index.js
```diff
@@ -0,0 +1,191 @@
...
+    }
+    containerObserver.current = new IntersectionObserver(
+      observerCb,
+      observerOptions
+    )
+
+    const first = childrenRef.current[0]
+    if (first) containerObserver.current.observe(first)
+
+    const last = childrenRef.current.pop()
```
quaardvark, line 99: This mutates the array. A big no-no

src/components/organisms/Testimonials/index.js
```diff
@@ -0,0 +1,191 @@
...
+          </Heading>
+        )}
+        <CarouselBox
+          ref={scrollRef}
+          mb="1rem"
+          mx={`-${scrollSpace}px`}
+          px={`${scrollSpace}px`}
+        >
+          <TestimonialsStack
+            wrapChildren
```
quaardvark, line 142: Now I understand, `position: relative` is missing here and `offsetLeft` is calculated from `window` which causes all kinds of chaos in calculations.

We need to add `position: relative` and remove `scrollSpace` from calculations of scrolls

src/components/organisms/Testimonials/index.js
```diff
@@ -0,0 +1,191 @@
...
+    if (last) containerObserver.current.observe(last)
+  }, [containerRef.current, childrenRef.current.length])
+
+  const scrollLeft = useCallback(() => {
+    const leftClipped = childrenRef.current
+      .slice()
+      .reverse()
+      .find(({ offsetLeft }) => offsetLeft - scrollRef.current.scrollLeft < 0)
+    if (leftClipped) {
+      scrollRef.current.scrollTo(scrollSpace - leftClipped.offsetLeft, 0)
```
quaardvark, line 112: I don't understand the need for `scrollSpace` here

src/components/organisms/Testimonials/index.js
```diff
@@ -0,0 +1,191 @@
...
+    }
+  }, [scrollSpace, scrollRef.current, childrenRef.current.length])
```
quaardvark, line 114: There should be no need to use `current` and `length` in dependencies here. This hook only really cares about non-stale references to `childrenRef` and `scrollRef`, not their values

src/components/organisms/Testimonials/index.js
```diff
@@ -0,0 +1,191 @@
...
+    )
+
+    const first = childrenRef.current[0]
+    if (first) containerObserver.current.observe(first)
+
+    const last = childrenRef.current.pop()
+    if (last) containerObserver.current.observe(last)
+  }, [containerRef.current, childrenRef.current.length])
+
+  const scrollLeft = useCallback(() => {
```
quaardvark, line 103: I reworked scrollLeft as such and thought it's a way better solution:
```
const scrollLeft = useCallback(() => {
  const leftClipped = childrenRef.current.find(
    ({ offsetLeft, offsetWidth }) =>
      offsetLeft + offsetWidth - scrollRef.current.scrollLeft >
      0 - containerRect.width
  )
  if (leftClipped) {
    scrollRef.current.scrollTo(leftClipped.offsetLeft, 0)
  }
}, [scrollRef, childrenRef])
```

Bottom line, we want left and right scrolls to scroll approximately same amount

src/components/organisms/Testimonials/index.js
```diff
@@ -0,0 +1,191 @@
...
+                    />
+                  )}
+                  <h5>{author}</h5>
+                  <span>{titleAndLocation}</span>
+                </Testimonial>
+              )
+            )}
+          </TestimonialsStack>
+        </CarouselBox>
+        <ScrollControlsContainer>
```
quaardvark, line 175: Should these hide when `!mayScrollLeft && !mayScrollRight`?

src/components/organisms/Testimonials/testimonials.stories.js
```diff
@@ -0,0 +1,46 @@
...
+    <Text>
+      Lorem ipsum dolor sit amet consectetur adipisicing elit. Expedita possimus
+      officiis tenetur, ex tempora repellat quibusdam, ab soluta neque incidunt
+      magni earum architecto veritatis rerum optio consectetur adipisci minus
+      dicta.
+    </Text>
+  )
+}
+
+export function Default() {
```
quaardvark, line 18: I think it would be good to add a few stories:

1) With 2 cards
2) With 12 cards

quaardvark (changes requested):

pkrawc:
@henriquecfreitas I just added a small (not exactly working) update to the testimonials layout as a story. It's not a slice and shouldn't affect anything about the slice you're currently building. Tomorrow morning let's go over the update after the stand with Nellie.

henriquecfreitas:
> @henriquecfreitas I just added a small (not exactly working) update to the testimonials layout as a story. It's not a slice and shouldn't affect anything about the slice you're currently building. Tomorrow morning let's go over the update after the stand with Nellie.

Ok!

src/components/organisms/Testimonials/index.js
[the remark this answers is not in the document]
quaardvark: @henriquecfreitas we should do that. If it's still unclear what needs to be done, - let me know

quaardvark:
@henriquecfreitas I think the observer cleanup function is still missing, let's add that and we should be good for QA

src/components/organisms/Testimonials/index.js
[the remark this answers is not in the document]
henriquecfreitas: Done!

PR #36, opened 2020-12-31
Add Prismic labels support for RichText, set default heading sizing for components

quaardvark:
**Resolves**

[Airtable - Separate Font Size from Heading](https://airtable.com/tbl4CgbPGPmsSc1pl/viwkpIRDfH7ure4zB/recf0AtqFMUD3zrtn)

**Why**

- Semantic headings hierarchy is important for the SEO, but in practice this hierarchy rarely matches the visual one. This call for separation of semantic H1...H5 from influencing the visual presentation of these headings. In other words, in actual CMS circumstances we might need to have H2 that look like H5

**Solution**

- Set default heading sizing for components/slices that have explicit headings, disregarding the tag itself
- Allow content editors to override default component sizing, as well as default sizing of Rich Text headers via Prismic Labels

**Also Included**

- Switches all headings usage to `<Box as="h{x}" />` to gain acess to `sx`
- Various tweaks and adjustments of spacing
- Switched `babel-eslint` to `@babel/eslint-parser` for optional chaining support, - there are some anecdotes that Babel parser is no longer needed for ESLint, but have no time to verify
- Fixes messy FAQ slice markup

**Screenshots**
- Prismic labels
Screenshot_20210105_144437

**Notes**

Regarding Prismic labels format and corresponding config entries: the goal was to make it a bit more idiomatic for CMS users, as well as abstract it a little from the theme / design system specifics, keeping it open for extension.

### DIFF
--- a/src/components/organisms/FAQ/faq.stories.js
+++ b/src/components/organisms/FAQ/faq.stories.js
@@ -1,19 +1,20 @@
-import React, { Fragment } from "react"
+import React from "react"
 import { Text } from "@rent_avail/typography"
+import { Box } from "@rent_avail/layout"
 import { FAQ } from "./index"
 
 export default { title: "Components/FAQ" }
 
 function Answer() {
   return (
-    <>
+    <React.Fragment>
       <Text>
         Lorem ipsum dolor sit amet consectetur adipisicing elit. Expedita
         possimus officiis tenetur, ex tempora repellat quibusdam, ab soluta
         neque incidunt magni earum architecto veritatis rerum optio consectetur
         adipisci minus dicta.
       </Text>
-    </>
+    </React.Fragment>
   )
 }
 
@@ -26,9 +27,14 @@ export function Default() {
   ]
   return (
     <FAQ
-      eyebrow="FAQ"
-      title="Answers to common questions"
-      description="Ab soluta neque incidunt magni earum architecto veritatis rerum optio consectetur adipisci minus dicta."
+      eyebrow={<Text>FAQ</Text>}
+      title={<Box as="h2">Answers to common questions</Box>}
+      description={
+        <Text>
+          Ab soluta neque incidunt magni earum architecto veritatis rerum optio
+          consectetur adipisci minus dicta.
+        </Text>
+      }
       questions={questions}
       py="4rem"
     />
--- a/src/components/partials/SliceZone/components/FrequentlyAskedQuestionsSlice/index.js
+++ b/src/components/partials/SliceZone/components/FrequentlyAskedQuestionsSlice/index.js
@@ -3,15 +3,9 @@ import { FAQ } from "components/organisms/FAQ"
 import { CONTAINER_WIDTHS } from "config"
 import RichText from "../RichText"
 
-const FrequentlyAskedQuestionsSlice = ({ slice }) => {
+export default function FrequentlyAskedQuestionsSlice({ slice }) {
   const {
-    primary: {
-      background,
-      color,
-      title,
-      description,
-      eyebrow,
-    },
+    primary: { background, color, title, description, eyebrow },
   } = slice
 
   const questions = slice.items.map(({ question, answer }) => ({
@@ -21,15 +15,14 @@ const FrequentlyAskedQuestionsSlice = ({ slice }) => {
 
   return (
     <FAQ
-      title={<RichText render={title} heading />}
-      description={<RichText render={description} />}
-      eyebrow={<RichText render={eyebrow} />}
+      title={title[0]?.text && <RichText render={title} />}
+      description={description[0]?.text && <RichText render={description} />}
+      eyebrow={eyebrow[0]?.text && <RichText render={eyebrow} />}
       questions={questions}
       containerWidth={CONTAINER_WIDTHS}
       bg={background}
       color={color}
+      py="6rem"
     />
   )
 }
-
-export default FrequentlyAskedQuestionsSlice
--- a/src/components/partials/SliceZone/components/FrequentlyAskedQuestionsSlice/index.js
+++ b/src/components/partials/SliceZone/components/FrequentlyAskedQuestionsSlice/index.js
@@ -15,9 +15,9 @@ export default function FrequentlyAskedQuestionsSlice({ slice }) {
 
   return (
     <FAQ
-      title={title[0]?.text && <RichText render={title} />}
-      description={description[0]?.text && <RichText render={description} />}
-      eyebrow={eyebrow[0]?.text && <RichText render={eyebrow} />}
+      title={title?.[0]?.text && <RichText render={title} />}
+      description={description?.[0]?.text && <RichText render={description} />}
+      eyebrow={eyebrow?.[0]?.text && <RichText render={eyebrow} />}
       questions={questions}
       containerWidth={CONTAINER_WIDTHS}
       bg={background}
--- a/src/components/partials/SliceZone/components/HowItWorksSlice/index.js
+++ b/src/components/partials/SliceZone/components/HowItWorksSlice/index.js
@@ -1,7 +1,7 @@
 import React from "react"
 import Anchor from "components/elements/Anchor"
 import { HowItWorks } from "components/organisms/HowItWorks"
-import { CONTAINER_WIDTHS, Typography } from "config"
+import { CONTAINER_WIDTHS, STYLING } from "config"
 import RichText from "../RichText"
 
 const HowItWorksSlice = ({ slice }) => {
@@ -13,7 +13,7 @@ const HowItWorksSlice = ({ slice }) => {
     uid: title?.[0]?.text || idx,
     copy: (
       <React.Fragment>
-        <RichText render={title} mb="2rem" sx={{ ...Typography.H3 }} />
+        <RichText render={title} mb="2rem" sx={{ ...STYLING.title }} />
         <RichText render={description} />
       </React.Fragment>
     ),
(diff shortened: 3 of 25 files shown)


src/components/partials/SliceZone/components/RichText/components/Text/index.js
```diff
@@ -0,0 +1,27 @@
+import React from "react"
+import styled from "styled-components"
+import { sx as applySx } from "@rent_avail/base"
+import { Text as _DesignSystemText } from "@rent_avail/typography"
+
+const DesignSystemText = styled(_DesignSystemText)`
+  ${applySx};
```
henriquecfreitas, line 7: This can only be done by creating a styled component? It could not be resolved via props?

src/components/partials/SliceZone/components/HowItWorksSlice/index.js
```diff
@@ -10,10 +10,10 @@ const HowItWorksSlice = ({ slice }) => {
   } = slice
   // eslint-disable-next-line no-shadow
   const sections = slice.items.map(({ title, description, image }, idx) => ({
-    uid: (title && title[0] && title[0].text) || idx,
+    uid: title?.[0]?.text || idx,
```
henriquecfreitas, line 13: Great job on this PR, optional chaining is awesome 😎

src/components/partials/SliceZone/components/FrequentlyAskedQuestionsSlice/index.js
```diff
@@ -21,15 +15,14 @@ const FrequentlyAskedQuestionsSlice = ({ slice }) => {
 
   return (
     <FAQ
-      title={<RichText render={title} heading />}
-      description={<RichText render={description} />}
-      eyebrow={<RichText render={eyebrow} />}
+      title={title[0]?.text && <RichText render={title} />}
+      description={description[0]?.text && <RichText render={description} />}
+      eyebrow={eyebrow[0]?.text && <RichText render={eyebrow} />}
```
henriquecfreitas, line 20: Why these properties are passed as arrays? There is any case where it could have more than one element on it?

henriquecfreitas (approved):

src/components/partials/SliceZone/components/RichText/components/Text/index.js
quaardvark: I guess you could, but why would you? This is the intended way to add `sx` prop support

src/components/partials/SliceZone/components/FrequentlyAskedQuestionsSlice/index.js
quaardvark: This is `StructuredText` type in Prismic, it will contain multiple elements for multi-paragraph entries

henriquecfreitas: So shouldn't this possibility be handled here? If `description` gets as a multi-paragraph text, we would be rendering only the first paragraph by accessing only the position `0`?
Think I understood it wrong

pkrawc: He's just checking to make sure at least one item in the array exists.

henriquecfreitas: Oh of course, my bad, sorry

pkrawc:
@quaardvark So you're proposing allowing the content creator to choose h1..5 as a heading when writing text, but then applying a label that really makes it a different number header?

If that's the case I think it's confusing and we should probably go with the labels named in the design system associated with size and weight, `hero`, `headline`, `title`, `subtitle`, `body--emphasis`, `body`, `small`.

quaardvark:
@pkrawc The idea is mostly to soft-lock styling of component headings to curated design decisions, disregarding which tag is passed from CMS. So for example, default `PitchCard` has H5 styled headings, but semantically it could/should be H3. So developers task is to ensure that headings are properly styled regardless of H level. In cases where editors need to override that - they can do that by labels. Naturally, this also extends to headings inside RichText, expect their default styling will be our default design system typography, but can also overridden with labels

On terminology, I was going to initially to set them as you suggest, but I thought that it could be very confusing for the CMS users, so I thought to keep it simple. I understand that these styles are not really headers-only and there's an opportunity to make it more generic, but I think we need to consider some middle ground between being meta and being comprehensive for CMS end-users. I also thought, that Typography interface will give an opportunity to abstract some other, more custom styling options to labels (like links, etc). 

Somewhat impactful problem with labels, is that there could only be one label per span, so going with very generic labels is kinda not possible, since we can't compose them.

pkrawc:
> @pkrawc The idea is mostly to soft-lock styling of component headings to curated design decisions, disregarding which tag is passed from CMS. So for example, default `PitchCard` has H5 styled headings, but semantically it could/should be H3. So the developer's task is to ensure that headings are properly styled regardless of H level. In cases where editors need to override that - they can do that by labels. Naturally, this also extends to headings inside RichText, expect their default styling will be our default design system typography, but can also be overridden with labels
> 
> On terminology, I was going to initially set them as you suggest, but I thought that it could be very confusing for the CMS users, so I thought to keep it simple. I understand that these styles are not really headers-only and there's an opportunity to make it more generic, but I think we need to consider some middle ground between being meta and being comprehensive for CMS end-users. I also thought that the Typography interface will give an opportunity to abstract some other, more custom styling options to labels (like links, etc).
> 
> Somewhat impactful problem with labels is that there could only be one label per span, so going with very generic labels is kinda not possible, since we can't compose them.

Sorry if I'm misunderstanding still, but the desired behavior is to coerce whatever heading that the CMS user chooses into the correct styling for text in the current slice?

I don't think we need labels for that, just use the asText method of the RichText component.

I still think of the labels as class names, and so they don't make much sense to me as h[N]. That seems like bad naming. I would also expect them to function like classes in that there's no point in putting two on a span where the job of both classes is to style the text differently (the one in the later position would override the previous class).

@quaardvark

quaardvark:
> 
> Sorry if I'm misunderstanding still, but the desired behavior is to coerce whatever heading that the CMS user chooses into the correct styling for text in the current slice?
> 
> I don't think we need labels for that, just use the asText method of the RichText component.

We don't, - we need labels to circumvent this behavior in cases where it's undesirable, so that the editor can still force H3 styling on otherwise H5-like by default component header, should they need to.

> 
> I still think of the labels as class names, and so they don't make much sense to me as h[N]. That seems like bad naming. I would also expect them to function like classes in that there's no point in putting two on a span where the job of both classes is to style the text differently (the one in the later position would override the previous class).
> 

These are not really class names, these are selection option in the dropdown for the CMS users. They are the intended user of labels, not us. Class names can be composed, these don't, so I feel like they could be a bit more specific. 

How would you otherwise propose to name the Heading 5 styling in a way that is not idiosyncratic to the design system and can be understood by external content management staff?

@pkrawc

pkrawc:
> How would you otherwise propose to name the Heading 5 styling in a way that is not idiosyncratic to the design system and can be understood by external content management staff?

I still think if the CMS user is going to have to learn a list of styling labels for text then we should just use the design system ones yeah? It's not like marketing is totally in the dark about design.avail.co or our brand guidelines. I think idiosyncrasy with the design system is actually probably desired here. KISS would indicate that multiple lists of styles for the dev and CMS user to keep in their heads is a recipe for disaster.

> These are not really class names, these are selection option in the dropdown for the CMS users. They are the intended user of labels, not us. Class names can be composed, these don't, so I feel like they could be a bit more specific.

Class names can be composed, but CSS cascades and overrides so assigning something like the following.
```css
.hero {
  font-size: 4rem;
  font-weight: bold;
}
.title {
  font-size: 3rem;
  font-weight: normal;
}
```
And "composing" them onto an element like `class="hero title"` would still only net you the title's styling.

quaardvark:
@pkrawc OK, I give up, I was 60:40 on this anyway. We would probably need some editor guide anyway and we can illustrate effects of these labels there.

pkrawc (changes requested):
Small comments, one of which is just a typo.

src/components/organisms/FAQ/faq.stories.js
```diff
@@ -1,19 +1,20 @@
...
 import { Text } from "@rent_avail/typography"
+import { Box } from "@rent_avail/layout"
 import { FAQ } from "./index"
 
 export default { title: "Components/FAQ" }
 
 function Answer() {
   return (
-    <>
+    <React.Fragment>
```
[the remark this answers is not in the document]
quaardvark: I actually prefer this way, since it stands out and instantly recognizable. The empty one is not recommended because sometimes you need `key={..}` in those

quaardvark:
@pkrawc renamed labels to conform with the design system, added `body` and `small` labels as well

src/components/organisms/FAQ/faq.stories.js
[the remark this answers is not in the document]
pkrawc: I don't disagree. I'm all in on using <Fragment/> instead of </> but you were using <React.Fragment/>.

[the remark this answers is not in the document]
quaardvark: Yeah, I prefer `<React.Fragment/>` over `<Fragment/>` for readability. If we want to change that let's enforce it via ESLint in a separate PR, since there are a few places it's used this way.

pkrawc (approved):
Thanks for these changes @quaardvark.

:shipit:

PR #37, opened 2021-01-12
Add Button/CTA Slice

henriquecfreitas:
Airtable item: https://airtable.com/tbl4CgbPGPmsSc1pl/viwQmvVkCTC16fDO3/recNyoXhzDRPIELtW?blocks=hide

Screenshots:
Screen Shot 2021-01-12 at 2 00 36 PM
Screen Shot 2021-01-12 at 2 00 04 PM
Screen Shot 2021-01-12 at 1 59 42 PM
Screen Shot 2021-01-12 at 1 59 30 PM
Screen Shot 2021-01-12 at 1 59 19 PM
Screen Shot 2021-01-12 at 1 57 06 PM
Screen Shot 2021-01-12 at 1 57 42 PM


src/components/partials/SliceZone/components/ButtonSlice/index.js
```diff
@@ -0,0 +1,32 @@
+import React from "react"
+import { ButtonCTA } from "components/organisms/ButtonCTA"
+import { CONTAINER_WIDTHS } from "config"
+
+const ButtonSlice = ({ slice }) => {
```
quaardvark, line 5: Should probably be called `ButtonCTASlice`? Could you also update content type in `types/info.json`?

src/components/organisms/ButtonCTA/index.js
```diff
@@ -0,0 +1,74 @@
...
+function ButtonCTA({
+  bg,
+  title,
+  buttonText,
+  buttonLink,
+  position,
+  containerWidth,
+  ...props
+}) {
+  const { flow, textAlign } = useMemo(() => {
```
quaardvark, line 17: I think this is great to implement with styled-system variants API - https://styled-system.com/variants

src/components/organisms/ButtonCTA/button.stories.js
```diff
@@ -0,0 +1,45 @@
+import React from "react"
+import { Box } from "@rent_avail/layout"
+import { ButtonCTA } from "./index"
+
+export default { title: "Components/Testimonials" }
```
quaardvark, line 5: `title` should be "Components/ButtonCTA"

src/components/organisms/ButtonCTA/index.js
```diff
@@ -0,0 +1,74 @@
...
+          textAlign: "left",
+        }
+    }
+  }, [position])
+
+  return (
+    <SkewBox bg={bg} {...props}>
+      <Container maxWidth={containerWidth} py="6rem">
+        <Flex sx={{ flexFlow: flow, gap: "2rem" }}>
+          <RichText
```
quaardvark, line 47: `organisms` shouldn't use `RichtText`, - it should be used in `...Slice`. `RichText` uses Prismic-specific DSL which should not leak into design components

src/components/organisms/ButtonCTA/button.stories.js
```diff
@@ -0,0 +1,45 @@
+import React from "react"
```
quaardvark, line 1: This storybook doesn't compile for me, because of the above issue with `RichText` being used in design component. Was it tested before opening a PR?

I also think adding a storybook with "No Title" variant would be great

quaardvark (changes requested):
Few small things, but also need to see updated `types/info.json`

src/components/organisms/ButtonCTA/index.js
henriquecfreitas: Great! Yes, its definitely better with it.
Thank you

src/types/info.json
```diff
@@ -87,6 +88,88 @@
...
+            "display" : "list",
+            "non-repeat" : {
+              "title" : {
+                "type" : "StructuredText",
+                "config" : {
+                  "single" : "heading1, heading2, heading3, heading4, heading5, heading6, strong, em",
+                  "allowTargetBlank" : true,
+                  "label" : "Title",
+                  "placeholder" : "Title",
+                  "labels" : {
```
quaardvark, line 105: More of a question, than feedback, - how did you manage to get this structure? Does that mean that we can use different values for name and value?

src/components/organisms/ButtonCTA/index.js
```diff
@@ -0,0 +1,82 @@
...
+              },
+            })}
+          {buttonLink &&
+            cloneElement(
+              buttonLink,
+              {},
+              <Button
+                flex="0"
+                variant="primary"
+                background={bg}
```
quaardvark, line 65: It seems like all the attributes after `variant` are unnecessary? I tried to remove them and couldn't see a difference

src/components/organisms/ButtonCTA/index.js
```diff
@@ -0,0 +1,82 @@
...
+                ...title.props?.sx,
+                flex: "1",
+                minWidth: "80%",
+              },
+            })}
+          {buttonLink &&
+            cloneElement(
+              buttonLink,
+              {},
+              <Button
```
quaardvark, line 62: Should use `as="a"` or `forwardedAs="a"` here, since `<button href=` is not a valid HTML, not good for users and SEO.

src/components/partials/SliceZone/components/ButtonCTASlice/index.js
```diff
@@ -0,0 +1,34 @@
...
+      color,
+      orientation = "left",
+      skew = "none",
+    },
+  } = slice
+
+  return (
+    <ButtonCTA
+      title={title && <RichText render={title} />}
+      buttonText={buttonText}
```
quaardvark, line 23: Could the whole button be implemented here and passed down, like in the other components? Instead of splitting it to text and link, which seems arbitrary. If it needs some specific positioning, it could be wrapped into something in design component.

quaardvark (changes requested):

src/types/info.json
```diff
@@ -87,6 +88,88 @@
           "testimonials" : [ ]
         },
         "choices" : {
+          "button" : {
```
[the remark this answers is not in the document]
henriquecfreitas: Is not, I'll rename it now

src/types/info.json
henriquecfreitas: I'm not sure about how I got it.
I think this might've been generated by the prismic interface...

Since we're not sure how it works, I'll replace it for the simple array, like in other slices

src/components/organisms/ButtonCTA/index.js
henriquecfreitas: The `type` and `display` are unnecessary, I'll remove it.
The `height`, `width` and `margin` attributes are being useful on specific orientations and screen sizes, to keep the button aligned and avoid its growth.
Now I'm not sure about the `textAlign`, but I'll test all of them again now

henriquecfreitas: `background ` is necessary to make the Button component set the right schema and colors.
Check it [here](https://github.com/rentalutions/marketing/blob/5cb73ebb11f363e435c1303b6634888fa8e683e7/src/components/elements/Button/index.js#L45)

src/types/info.json
quaardvark: I can't find how to put labels in Prismic UI, that's why I'm asking

src/components/organisms/ButtonCTA/index.js
quaardvark: Could we do the positioning via some kind of `<Box/>` instead? I think it would be great if we can allow the parent component to implement the "button". It could be useful for extension, - use some other things in place here.

henriquecfreitas: Yes, we can implement the Button in the Slice component.
Actually I think we don't even need the `<Box/>`, since we can set the positioning attributes directly on the button component received as prop from the parent, don't we?

src/types/info.json
henriquecfreitas: Got it...
Unfortunately I don't know how to answer it. This code block was copied from the one I was working on at the Prismic UI

src/components/partials/SliceZone/components/ButtonCTASlice/index.js
henriquecfreitas: Sure, I'm working on it

quaardvark:
@henriquecfreitas I cannot see this slice in Prismic anymore. Do you know what happened?

quaardvark (approved):

PR #41, opened 2021-01-19
Creating Plans/Prices slice

henriquecfreitas:


src/components/organisms/PlansPrices/index.js
```diff
@@ -0,0 +1,139 @@
+import React, { cloneElement } from "react"
+import styled from "styled-components"
+import { variant } from "styled-system"
+
+import { Box, Card, Container, Flex, Grid } from "@rent_avail/layout"
+import SkewBox from "components/molecules/SkewBox"
+import { STYLING } from "config"
+import PlanInfo from "./planInfo"
```
quaardvark, line 8: We should stick to kebab case for filenames, - `plan-info`, `plans-prices.stories`, etc

src/components/partials/SliceZone/components/PlansPricesSlice/index.js
```diff
@@ -0,0 +1,67 @@
...
+  )
+
+  return (
+    <PlansPrices
+      containerWidth={CONTAINER_WIDTHS}
+      bg={background}
+      color={color}
+      skew={skew}
+      direction={direction}
+      title={title && <RichText render={title} />}
```
quaardvark, line 52: IIRC Rich Text fields from Prismic are always non-null, so it might render empty tag here, unless it's what you want ofc

quaardvark (commented):
Some small thing, - overall looks fine

src/components/partials/SliceZone/components/PlansPricesSlice/index.js
henriquecfreitas: I think it would be better to pass it as no-value instead of an empty RichText, because of the margins that would be applied on the empty element, causing a bigger gap between the last and next visible elements
But I might be wrong, what do you think?

quaardvark: Headers are actually checked in `htmlSerializer` and should not render on empty, but it's safe to check here as well, I don't remember what's the exact structure though

quaardvark (approved):

PR #42, opened 2021-01-21
Slices fade in animation

henriquecfreitas:


src/components/organisms/ButtonCTA/index.js
```diff
@@ -35,6 +35,21 @@ const StyledFlex = styled(Flex)(
...
+//     // style
+//   },
+//   bottom: {
+//     // style
+//   },
+//   right: {
+//     // style
+//   }
+// }
+
```
pkrawc, line 51: You can likely remove this comment block unless you think you'll use it in the future.

src/components/organisms/ButtonCTA/index.js
```diff
@@ -43,12 +58,13 @@ function ButtonCTA({
...
   ...props
 }) {
+  console.log(button)
   return (
     <SkewBox bg={bg} {...props}>
       <Container maxWidth={containerWidth} py="6rem">
         <StyledFlex orientation={orientation}>
           {title &&
-            cloneElement(title, {
+            useFadeIn(title, {
```
pkrawc, line 65: Interesting. It's not a hook now I understand. We shouldn't be naming anything `useSomething` if we're not expecting it to be used as a hook.

pkrawc (changes requested):
Hey @henriquecfreitas,

Looks like you have a failing build. I started running through a review and I'll keep my comments below but some of them might be outdated. Let me know when you've reviewed my comments and I can begin another look through of this draft.

Thanks man!

src/components/fadeIn.js
```diff
@@ -0,0 +1,48 @@
...
+  ...props
+}) {
+  const [hidden, setHidden] = useState(true)
+  const [target, { isIntersecting }] = useIntersection(intersectionOptions)
+
+  useEffect(() => {
+    if (hidden) {
+      setHidden(!isIntersecting)
+    }
+  }, [isIntersecting])
```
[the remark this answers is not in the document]
henriquecfreitas: > Hey @henriquecfreitas,
> 
> Looks like you have a failing build. I started running through a review and I'll keep my comments below but some of them might be outdated. Let me know when you've reviewed my comments and I can begin another look through of this draft.
> 
> Thanks man!

Yeah, its failing, unfortunately. I'm trying to make it work, but this is taking more time than the expected.
I pushed this commit so you and @quaardvark can make some suggestions on how things are going on the code until now.
The storybook is working, if you want to have a look at the effects on each component

src/components/organisms/ButtonCTA/index.js
henriquecfreitas: I'll remove it and open another PR to promote that change

src/components/organisms/ButtonCTA/index.js
henriquecfreitas: I think I misunderstood what you suggested on the first time we talked about that `withFadeIn` HOC.
But given your hook example I think we should keep this version and rename it to something like `applyFadeIn`, because I think it will make things easier for us to change the components we are already have

src/components/fade-in.js
```diff
@@ -0,0 +1,64 @@
+import React, { useState, useEffect, forwardRef } from "react"
```
quaardvark, line 1: Probably should be in `elements` or `molecules`

src/components/partials/SliceZone/components/PitchCardsSlice/index.js
```diff
@@ -41,16 +44,18 @@ const PitchCardsSlice = ({ slice }) => {
...
-              button: linkAsButton,
-            }
-          : null,
+      isButtonVariant: linkAsButton,
+      link: linkText && linkUrl && (
+        <Link link={linkUrl} passHref>
+          {createElement(linkAsButton ? Button : Text, {
+            as: "a",
+            ...getTargetProps(
+              linkUrl.target || (linkUrl.link_type === "Media" && "_blank")
```
quaardvark, line 53: IIRC this logic is already present in `Link`. Could you verify?

src/components/organisms/Showcase/cases.js
```diff
@@ -0,0 +1,46 @@
+import React, { useEffect, useState } from "react"
```
quaardvark, line 1: I don't think breaking component up for the sake of breaking it up is a good idea, it makes it harder to read the `index.js`, not easier, - `<Cases />` is a needle in a haystack there

quaardvark (commented):
I don't really like the API in general. The whole project is now constructed of `<FadeIn as={Container} />` components, where `FadeIn` is a minor feature, not a building block. I think we are overusing `as=` API at this point, - makes it hard to read.

I think we should aim for solution where components can exist without `FadeIn`. Not necessarily because we would need that, but because this is a good constraint to build a good API. What if I would like another animation?

@pkrawc @henriquecfreitas Is there a way to wrap all the necessary elements in an abstract animation interface that can take animation style as a prop(s), or HOC option, - including not animating at all?

src/components/fade-in.js
henriquecfreitas: It is here temporarily. I think it should be on [elements repo](https://github.com/rentalutions/elements), am I right?

henriquecfreitas:
> I don't really like the API in general. The whole project is now constructed of `<FadeIn as={Container} />` components, where `FadeIn` is a minor feature, not a building block. I think we are overusing `as=` API at this point, - makes it hard to read.
> 
> I think we should aim for a solution where components can exist without `FadeIn`. Not necessarily because we would need that requirements wise, but because this is a good constraint to build a good API. What if I would like another animation?
> 
> @pkrawc @henriquecfreitas Is there a way to wrap all the necessary elements in an abstract animation interface that can take animation style as a prop(s), or HOC option, - including not animating at all?

Maybe creating a higher component that receives an optional `effect` property that can be `FadeIn` or any other effect component.
But I also think the HOC would be a good approach.

pkrawc:
> I don't really like the API in general. The whole project is now constructed of `<FadeIn as={Container} />` components, where `FadeIn` is a minor feature, not a building block. I think we are overusing `as=` API at this point, - makes it hard to read.
> 
> I think we should aim for a solution where components can exist without `FadeIn`. Not necessarily because we would need those, requirements wise, but because this is a good constraint to build a good API. What if I would like another animation?
> 
> @pkrawc @henriquecfreitas Is there a way to wrap all the necessary elements in an abstract animation interface that can take animation style as a prop(s), or HOC option, - including not animating at all?

I think animation states are fairly building block level architecture. We can consider them table stakes now. I also don't think we're overusing the `as` API, but can see how it would make it harder to read.

Components can easily exist without the `FadeIn`. Just use `motion.whatever` as your animation primitive, I'd be happy to transfer the logic back to framer motion but the FadeIn seems like a common and useful tool in a lot of scenarios--because it also tracks its children and if they're intersecting the viewport.

I do **not** like the idea of an HOC, it's an antipattern in my mind. It hides implementation logic, makes the component structure harder to refactor, and requires unnecessary function declarations to use at a granular level. Eg.

```jsx
const SmallPart = withFadeIn(SmallPart)

const OtherPart = withFadeIn(OtherPart)

function BigPart() {
  return (
    <Comp>
      <SmallPart />
      <OtherPart />
    </Comp>
  )
}

// Instead of

function BigPart() {
  return (
    <Comp>
      <FadeIn className="small-part" />
      <FadeIn className="other-part" />
    </Comp>
  )
}
```

quaardvark (commented):
I think we should move the hook to this repository, polish it out and ship it to elements later, - this unblocks us to review this PR.

Also, do we bail on orchestration, or is it still in the works? It's OK if we do, just curious what came out of it.

henriquecfreitas:
> I think we should move the hook to this repository, polish it out and ship it to elements later, - this unblocks us to review this PR.
OK

> Also I still see the delays sprinkled everywhere, - do we bail on orchestration, or is it still in the works? It's OK if we do, just curious what came out of it.
The orchestration props were overriding the top transition props, causing the children to animate very different from the other elements.
It can still be used if we set the attrs directly on the animated components, but I think that manually setting the delay and duration on each child is a easier way to go for this first release

src/components/organisms/ButtonCTA/index.js
```diff
@@ -43,19 +46,32 @@ function ButtonCTA({
   containerWidth,
   ...props
 }) {
+  const [ { fadeIn }, target ] = useAnimateIntersection()
```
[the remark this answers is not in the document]
henriquecfreitas: This way we can get different animations using the same hook (and consequently the same observer), and for each element apply the animation we want, or don't apply any animation at all.
` 
 const [ { fadeIn, fadeOut, scaleIn }, target ] = useAnimateIntersection()
return <MySlice ref={target}>
  <motion.aside  {...fadeIn}>
    <TopComponent />
  <motion.aside/>
  <motion.aside  {...fadeOut}>
    <MiddleComponent />
  <motion.aside/>
  <motion.aside>
    <NonAnimatedComponent />
  <motion.aside/>
  <motion.aside  {...scaleIn}>
    <BottomComponent />
  <motion.aside/>
</MySlice>
`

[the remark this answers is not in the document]
quaardvark: So the animation style is decided on the component internals level, and not on consumer level? This makes it kinda pointless, since it's virtually the same as hardcoding `<FadeIn>`, as in initial implementation. 

My suggestion was that the design component have abstract support for animation, but concrete animation preset is passed from the consumer, and that by default it has no animation whatsoever.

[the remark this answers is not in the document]
pkrawc: To get to what @quaardvark is suggesting. I at one point proposed the following:

```jsx
const [presets, ref] = useInViewAnimation() // I like this name btw. Very clear.
return <motion.section {...presets[preset] } /> // Preset comes from props.
```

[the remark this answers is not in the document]
henriquecfreitas: I would like to keep the hook the way it is, because it gives us the freedom to use it on different ways if, for example, we want to place different animations on the same slice/component.

In order to not having to place the `{...fadeIn}` in every element that will be animated I would go with something like what suggest @pkrawc suggested:
```js
const [presets, ref] = useInViewAnimation()
const animation = presets[animationPreset]

return <motion.section {...animation} />
```

henriquecfreitas:
@pkrawc when you pushed the merge you also pushed 2 obsolete commits that weren't on the commit tree anymore (f2c0124 and 6e61a3a). Was it intentional or they were pushed by mistake?

pkrawc:
@henriquecfreitas No, I didn't mean to push those after orphaned. This is a good reason not to squash commits. We shouldn't be squashing anything honestly. An interactive git merge at the end of a PR life cycle could be useful to clean up commits but removing commits on the tree while multiple people are working on something will always result in something like this happening.

quaardvark:
@pkrawc @henriquecfreitas I pushed an mini-overhaul, which breaks down to following:

- Simplified the hooks in the animation utils, moved them to the utils folder
- Removed all manual delays and durations, - duratons are now default from Framer. They feel right to me, LMK if we need to adjust
- Added the orchestration setup: now `useAnimation` returns an object of shape `{ container: ..., item: ... }` where `container` holds the animation props for container itself (not animating by itself, just children stagger settings) and `item` holds the animation props for children. Note that children of children would not animate / stagger properly, - you would need to "reset" a `container` for that
- Some general fixes, simplifications

There are still a few things to fix / improve, so keeping it as a draft for now, but want your feedback nevertheless

The MegaPage for testing:
https://marketing-git-feature-fade-in-animation.rent-avail.vercel.app/info/listings-dev-test

src/utils/animation.js
```diff
@@ -0,0 +1,110 @@
+import { useRef, useEffect, useState, useCallback } from "react"
+
+export function useHadIntersected({ threshold } = {}) {
```
henriquecfreitas, line 3: This hooks will be moved to elements repo?
If we're going to keep them here, I would create a separated file for each hook. Especially `useHadIntersected` that is not necessarily related to animations

henriquecfreitas, line 3: Shouldn't we be setting a default value for threshold here?
I think that a higher value here would be acceptable, but we should keep `10%` as the default value in useInViewAnimation

src/utils/animation.js
```diff
@@ -0,0 +1,110 @@
...
+  itemDuration,
+  staggerChildren = 0.15,
+  staggerDirection = 1,
+  when,
+} = {}) {
+  const [final, setFinal] = useState(false)
+
+  function animate() {
+    setFinal(true)
+  }
```
pkrawc, line 36: Should this allow you to toggle between true and false?

pkrawc (approved):
I think I grok the whole thing and it definitely seems cleaner than the mess I left Henrique with. Wondering if it's useful to give the consumer the ability to "de-animate" the component ever.

But other than that small possible change, looks good to me.

src/components/organisms/PlansPrices/plan-card.js
```diff
@@ -0,0 +1,132 @@
...
+  buttonColor,
+  buttonBackground,
+  background,
+  color,
+  direction,
+  animationPreset = "fadeIn",
+}) {
+  const [presets, intersectionView] = useInViewAnimation({
+    containerDuration: "default",
+    threshold: 0.25,
```
henriquecfreitas, line 49:
```suggestion
```
Now this threshold is higher than the default, so I think we don't need to set it anymore

src/utils/animation.js
pkrawc: The browser sets a default threshold, so it's not required that we define it.

henriquecfreitas: Yes, but the default is `0`, so if a single pixel of the element is visible, it will be considered as intersected.
 I'm not saying that this would be wrong (actually is right, because if a pixel of the element is intersecting, then the entire element is intersecting as well), but is this what we want? Considering the possible proposes of this hook..

src/components/organisms/PlansPrices/plan-card.js
quaardvark: This threshold is deliberately higher, because it's a smaller / contained element.

src/utils/animation.js
quaardvark: There is no reason to pass any value here, default browser's is exactly what any engineer would expect. You could argue about `0`, but a higher value doesn't make any sense

quaardvark: I expect to have an extension to elements `useIntersection`, after which this file will be an animation utility, I don't think it's useful to split it just yet.

src/utils/animation.js
quaardvark: You know I hate toggles, so I just exposed `setFinal` to the consumer

quaardvark:
Also added explicit `"none"` preset, this was kinda supported with any invalid key, but better to have it there for visibility

src/utils/animation.js
pkrawc: Haha, yeah. I was thinking that would be the best solution as well. Glad we're on the same page.

PR #43, opened 2021-01-22
Add video support to Hero, How it Works adn Pitch Cards slices

quaardvark:


quaardvark:
@pkrawc @henriquecfreitas I'm merging it now for the demo, but still think it's worth reviewing

PR #44, opened 2021-01-28
Creating new testimonials slice

henriquecfreitas:


henriquecfreitas:
@pkrawc will we use this "background beyond the container" thing again on for new slices? (or for revisions of current slices, etc)
Because I think this is something worthy to be developed as a reusable component, like [SkewBox](https://github.com/rentalutions/marketing/blob/master/src/components/molecules/SkewBox/index.js)

src/components/organisms/Testimonials/v2/index.js
```diff
@@ -0,0 +1,182 @@
...
+        const nextIndex = getSafeIndex(activeIndex + 1)
+        setActiveIndex(nextIndex)
+      }, currentInterval * 1000)
+      return () => {
+        clearInterval(timeout)
+      }
+    }
+  }, [currentInterval, testimonials, activeIndex])
+
+  const Picture = useCallback((testimonialIndex, level = 0) => {
```
pkrawc, line 71: I love the creative use of `useCallback` here but there's actually a part of the React API specifically designed for this and it's `React.memo`

```jsx
const Picture = memo(function Picture(props) {}, [optionalCompareFunction])
```

src/components/organisms/Testimonials/v2/index.js
```diff
@@ -0,0 +1,182 @@
...
+              justifyContent="center"
+              sx={{
+                gap: 3,
+              }}
+            >
+              {Picture(getSafeIndex(activeIndex - 2), 2)}
+              {Picture(getSafeIndex(activeIndex - 1), 1)}
+              {Picture(activeIndex)}
+              {Picture(getSafeIndex(activeIndex + 1), 1)}
+              {Picture(getSafeIndex(activeIndex + 2), 2)}
```
pkrawc, line 181: Why not just use these as standard components? This is likely related to my `React.memo` comment above

```jsx
// Example
<Picture index={getSafeIndex(idx - 2)} key={2} />
```

pkrawc (changes requested):

quaardvark:
I'm not sure `v1` and `v2` are good names for these. Maybe something like `TestimonialsCards` and `TestimonialsCarousel` would be better? Don't you think @pkrawc ?

src/components/organisms/Testimonials/v2/index.js
henriquecfreitas: Great approach! Thank you for the tip

quaardvark:
@pkrawc @henriquecfreitas What do you guys think about the content type for this? Two slices or one slice with a `variant` field? How likely is that their data models will diverge eventually?

I kinda feel like even if data itself will be largely the same, potential customization options can vary, as well as their effect, which could lead to some end-user confusion in case of one slice. Like one having a skew, but not another, etc. So my gut feeling says two slices right now, but I'm not overconfident about that.

src/components/molecules/TitleOverflowContainer/index.js
```diff
@@ -0,0 +1,72 @@
+import React, { useEffect, useRef, useState } from "react"
+import { Box, Container, Flex } from "@rent_avail/layout"
+import { useResize } from "@rent_avail/utils"
+
+function TitleOverflowContainer({
```
quaardvark, line 5: I'm not sure this name is useful. Could you explain what it is and can it be used standalone or in other slices?

henriquecfreitas: It can be used in other slices. (I don't think that would make sense to use it standalone)
This component is meant to hold the container content on half of its content, and place a large title on the other half. It can be oriented to the right or to the left, and should break in two rows (title above, content bellow) when it doesn't fit the screen.
I created it as a molecule because I think that it should be reused by other slices organisms when there's need for it (like the SkewBox component)

henriquecfreitas:
> @pkrawc @henriquecfreitas What do you guys think about the content type for this? Two slices or one slice with a `variant` field? How likely is that their data models will diverge eventually?
> 
> I kinda feel like even if data itself will be largely the same, potential customization options can vary, as well as their effect, which could lead to some end-user confusion in case of one slice. Like one having a skew, but not another, etc. So my gut feeling says two slices right now, but I'm not overconfident about that.

I think both approaches are OK, creating a new slice component or using the same slice component to handle props and then call different organisms.

I created it are using the same slice component because most of the props are the same, and they both share the same handler for the repeatable zone items. Also because I thought that would be better for the organization of the code to keep them near each other, in a single Testimonials "package" that exports both versions. The component that imports it have the responsibility to decide which one to use on each case.


> Like one having a skew, but not another, etc

Actually this is totally possible right now, because the Skew is defined on the organism component, not on the slice level. So we can remove the skew from the Carousel version without removing it from the Cards. Same goes for the `TitleOverflowContainer` that is defined right bellow the `SkewBox` on the Carousel, but is not used by the Cards.

src/components/molecules/TitleOverflowContainer/index.js
quaardvark: That's fine if it's needed, but I still can't picture it from the name. Is this a title or a container for title? Is the title overflowing or container? How will another developer understand that he's supposed to reuse that, especially w/o the storybook?

henriquecfreitas: Is a Container that has a Title Box overflowing its margins.
I spent some time trying to think of a name for this component and I'm also not 100% positive with this name.
If you have any suggestions, it would be great.

About the storybook, I'll create it for this component and for the SkewBox

quaardvark: Yeah, overflow is an implementation detail. It can be done without overflow as well, so it shouldn't be part of a name. How about `BoxedTitleSection`? Ideally we could also try to split out the `BoxedTitle` child that is more generic, maybe we would want to use that style somewhere else, but we could do it later

henriquecfreitas: I like the `BoxedTitleSection` name, let use it.
But be aware that it can **not** be done without overflow. The box overflowing the container is the main propose of this component.

Screen Shot 2021-02-01 at 11 23 14 AM

quaardvark: I'm not saying that it **should** be done w/o overflow, but it's 100% possible.

src/components/organisms/Testimonials/carousel/index.js
```diff
@@ -0,0 +1,173 @@
...
+  const getSafeIndex = useCallback(
+    (desiredIndex) => {
+      const { length } = testimonials
+      const remainder = desiredIndex % length
+      return remainder < 0 ? remainder + length : remainder
+    },
+    [testimonials]
+  )
+
+  const Picture = memo(function Picture({ testimonialIndex, level = 0 }) {
```
pkrawc, line 43: This should be moved outside of the component unless there's something I'm not getting.

pkrawc (changes requested):
Sorry for the radio silence while I was pulled into other projects.

I've left a couple of pieces of feedback to clean up some of the code.

src/components/molecules/BoxedTitleSection/index.js
```diff
@@ -0,0 +1,84 @@
...
+
+  useEffect(() => {
+    const container = containerRef.current
+    if (!container || !getComputedStyle) return
+    const { marginLeft, paddingLeft } = getComputedStyle(container)
+    const titleSpaceInPixels = overflow
+      ? parseInt(marginLeft, 10) + parseInt(paddingLeft, 10)
+      : 0
+    setTitleSpace(`${titleSpaceInPixels}px`)
+  }, [containerRect])
```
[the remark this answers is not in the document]
henriquecfreitas: I had tried before in some other ways, this was the best approach that I could get that resulted in the expected margin and without overflowing the viewport. I'm open to suggestions.
Making components overflow the container is complicated, we had the same problem on the cards slide of the first Testimonials version.

src/components/organisms/Testimonials/cards/index.js
```diff
@@ -0,0 +1,202 @@
...
+  flex: 1 0 auto;
+  width: 26rem;
+  height: 100%;
+  display: flex;
+  flex-flow: column;
+  align-items: center;
+  justify-content: center;
+  text-align: center;
+  border-radius: 12px;
+`
```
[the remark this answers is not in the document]
henriquecfreitas: I agree that your suggestions on the `Testimonials/cards` are valid and that we should discuss them, but I don't think that this is the proper PR to do it. This file is exactly the same that is currently versioned at [Testimonials/index.js](https://github.com/rentalutions/marketing/blob/master/src/components/organisms/Testimonials/index.js), I just migrated it to a sub-directory in order to separate the `Cards` and the `Carousel` components, leaving the `index.js` responsible only to export them

src/components/organisms/Testimonials/cards/index.js
```diff
@@ -0,0 +1,202 @@
...
+
+function Testimonials({
```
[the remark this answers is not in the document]
henriquecfreitas: Same as https://github.com/rentalutions/marketing/pull/44#discussion_r570282527

src/components/molecules/BoxedTitleSection/index.js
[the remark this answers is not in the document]
quaardvark: Something like that?
https://codesandbox.io/s/overflow-wo-overflow-703tv?file=/src/index.js

src/components/organisms/Testimonials/carousel/picture.js
```diff
@@ -0,0 +1,62 @@
+import React from "react"
+import { Box } from "@rent_avail/layout"
+
+function Picture({
```
quaardvark, line 4: Let's give it a both more specific and more generic name, like `TestimonialsCarouselItem`
1) I can see it being not a picture
2) Non-derived names in specific / one-time components can be confused for some generic component

src/components/organisms/Testimonials/carousel/index.js
```diff
@@ -0,0 +1,145 @@
+import React, { cloneElement, useCallback, useEffect, useState } from "react"
+import { Box, Flex } from "@rent_avail/layout"
+import { Text } from "@rent_avail/typography"
+import SkewBox from "components/molecules/SkewBox"
+import BoxedTitleSection from "components/molecules/BoxedTitleSection"
+import { STYLING } from "config"
+import Picture from "./picture"
+
+function Testimonials({
```
quaardvark, line 9: Let's give both components a full name for proper/better IDE function and search

src/components/organisms/Testimonials/carousel/index.js
```diff
@@ -0,0 +1,145 @@
...
+          setActiveIndex(nextIndex)
+        }
+      }, currentInterval * 1000)
+      return () => {
+        clearInterval(timeout)
+      }
+    }
+  }, [currentInterval, testimonials, activeIndex])
+
+  const carouselItems = [
```
quaardvark, line 53: I wonder if this pattern will work should we decide to animate the items, - it currently works as a bit shift, but what if we would want to actually move/scale them. Not sure if it's worth thinking about it right now @henriquecfreitas @pkrawc

quaardvark (changes requested):

quaardvark:
@henriquecfreitas we need to resolve conflicts here and also add slice-level animation support

src/components/organisms/Testimonials/carousel/index.js
pkrawc: @henriquecfreitas @quaardvark If we keep the key constant for the item, framer-motion can handle the shift in an animated way. You'd want to probably use the index + a page count.

src/components/organisms/Testimonials/cards/index.js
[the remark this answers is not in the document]
pkrawc: Gotcha, I got lost in the versions of this slice and naming conventions.

src/components/molecules/BoxedTitleSection/index.js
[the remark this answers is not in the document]
henriquecfreitas: @pkrawc I couldn't make it work properly with `position: absolute`, this issue comes since the other version of the Testimonials, where we had a similar problem and @quaardvark also suggested the absolutely positioned container approach. There are too many thing to handle in order to make it work the way it should. I think the negative margin approach is easier for handling this.

@quaardvark your suggestion does works, but I honestly didn't appreciated using that `linear-gradient` for achievement the same result we can archive with the negative margins. It sounds for me more as a work-around rather than a concrete and scalable solution.

henriquecfreitas:
@pkrawc @quaardvark I just rebased the PR with staging, applied the animations rules on the TestimonialsCarousel, fixed some bugs and applied some suggestions from both of you.
I resolved all the applied/obsolete change requests here. Some other topics are still open, we'll need to discuss them again, can you take a look on the open change requests here?

src/components/molecules/BoxedTitleSection/index.js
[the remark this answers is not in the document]
quaardvark: Well, I'm not sure it's fair to say that negative margin and JS calculation is less "workaroundy" and more concrete than `linear-gradient`, but let's say it is.  

What about the following kind of layout?
```
+--------+-----------------+-----------------+--------+
|  Flex  |  Container / 2  |  Container / 2  |  Flex  |
+--------+-----------------+-----------------+--------+
```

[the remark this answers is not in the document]
quaardvark: https://codesandbox.io/s/overflow-wo-overflow-forked-sf3qj?file=/src/index.js

src/components/organisms/Showcase/index.js
```diff
@@ -72,9 +72,9 @@ const Showcase = ({
               m="2rem 0 3rem"
               minHeight="4.5rem"
             >
-              {cases.map(({ icon, copy }, idx) => (
+              {cases.map(({ icon, copy: caseCopy }, idx) => (
                 <Box
-                  key={copy}
+                  key={caseCopy}
                   color={idx === activeCase ? "blue_500" : "ui_500"}
```
pkrawc, line 78: Is this a workaround for `:hover`? I know this wasn't changed for this PR but just wondering.

src/components/organisms/Tabs/index.js
```diff
@@ -43,16 +43,15 @@ function useTabs({ defaultIndex, searchParam, ...htmlProps }) {
       const index = prevTabs.findIndex((tabEl) => {
         if (!tabEl.element || !element) return false
         return Boolean(
-          tabEl.element.compareDocumentPosition(element) &
+          tabEl.element.compareDocumentPosition(element) &&
```
pkrawc, line 46: This is a bitwise operation, shouldn't be a double ampersand.

src/components/organisms/Tabs/index.js
```diff
@@ -213,11 +191,11 @@ export function TabPanels(props) {
...
     React.isValidElement(child)
   )
-  const children = validChildren.map((child, idx) => {
-    return cloneElement(child, {
+  const children = validChildren.map((child, idx) =>
+    cloneElement(child, {
       isSelected: idx === selectedIndex,
     })
-  })
+  )
```
pkrawc, line 198: This doesn't read more clearly to me. Braces around function expressions should be used unless one line.

src/components/organisms/Tabs/index.js
```diff
@@ -137,22 +131,6 @@ function useTab({
...
-    return cloneElement(child, {
-      isSelected: idx === selectedIndex,
-    })
-  })
-  return {
-    ...props,
-    children,
-  }
-}
-
```
pkrawc, line 155: Did we just get rid of this hook entirely?

src/components/organisms/Testimonials/cards/index.js
```diff
@@ -0,0 +1,228 @@
+import React, {
```
pkrawc, line 1: 1. All the logic for testimonials should be moved to a portable `useTestimonials` hook.
2. The way styled-components are declared here is overkill. Prefer the `sx` prop.

src/components/organisms/Testimonials/cards/index.js
```diff
@@ -0,0 +1,228 @@
...
+
+  const containerRect = useResize(containerRef)
+
+  const [scrollSpace, setScrollSpace] = useState(0)
+  const [mayScrollLeft, setMayScrollLeft] = useState(false)
+  const [mayScrollRight, setMayScrollRight] = useState(false)
+
+  useEffect(() => {
+    const container = containerRef.current
+    if (!container || !getComputedStyle) return
```
pkrawc, line 81: Probably don't need to check for getComputedStyle in a useEffect?

src/components/organisms/Testimonials/carousel/testimonials-carousel-item.js
```diff
@@ -0,0 +1,62 @@
...
+  level = 0,
+  sx,
+  ...props
+}) {
+  const [opacity, scale] = (() => {
+    if (level >= 3) return [0.9, 0.5625]
+    if (level === 2) return [0.75, 0.625]
+    if (level === 1) return [0.5, 0.75]
+    return [0, 1]
+  })()
```
pkrawc, line 19: We can probably use `framer-motion`'s `useTransform` for this.

```jsx
const motionLevel = useSpring(level)
const opacity = useTransform(motionLevel, [0,3], [0.5, 1])
const scale = useTransform(motionLevel, [0, 3], [0.5, 0.75])
useEffect(() => {
  motionLevel.set(level)
}, [level])
```

src/components/organisms/Testimonials/carousel/index.js
```diff
@@ -0,0 +1,159 @@
+import React, { cloneElement, useCallback, useEffect, useState } from "react"
```
pkrawc, line 1: Logic should be extracted out to a hook.

src/components/partials/SliceZone/components/TestimonialsSlice/index.js
```diff
@@ -1,29 +1,46 @@
...
-import { CONTAINER_WIDTHS, STYLING } from "config"
+import {
+  TestimonialsCards,
+  TestimonialsCarousel,
+} from "components/organisms/Testimonials"
+import { CONTAINER_WIDTHS } from "config"
 import RichText from "../RichText"
 
-const TestimonialsSlice = ({ slice }) => {
+const TestimonialsSlice = ({ version = "cards", slice }) => {
```
pkrawc, line 9: I don't think we should be defaulting to the card version.

This brings up a larger question I have. We're building these testimonials in tandem, but I was under the assumption that the carousel was **replacing** the cards version.

src/components/partials/SliceZone/components/TestimonialsSlice/index.js
```diff
@@ -1,29 +1,46 @@
...
 
+  const Testimonials = (() => {
+    switch (version) {
+      case "carousel":
+        return TestimonialsCarousel
+      case "cards":
+      default:
+        return TestimonialsCards
+    }
+  })()
```
pkrawc, line 42: I really dislike IIFEs. Is there another way to write this?

src/components/partials/SliceZone/index.js
```diff
@@ -36,12 +36,15 @@ const SliceZone = ({ slices }) => {
...
+      case "plans_and_prices":
         return <PlansPricesSlice key={key} slice={slice} />
       case "show_case":
         return <ShowcaseSlice key={key} slice={slice} />
       case "testimonials":
-        return <TestimonialsSlice key={key} slice={slice} />
+      case "testimonials_cards":
+        return <TestimonialsSlice key={key} slice={slice} version="cards" />
+      case "testimonials_carousel":
+        return <TestimonialsSlice key={key} slice={slice} version="carousel" />
```
pkrawc, line 47: Seems like we're duplicating checks here, no?

pkrawc (changes requested):

src/components/organisms/Showcase/index.js
quaardvark: `activeCase` is set by clicks and interval god

src/components/elements/Anchor/index.js
```diff
@@ -6,8 +6,6 @@ const StyledAnchor = styled.a`
   display: block;
 `
 
-const Anchor = ({ hash }) => {
-  return <StyledAnchor name={hash} id={hash} />
-}
+const Anchor = ({ hash }) => <StyledAnchor name={hash} id={hash} />
```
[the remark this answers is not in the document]
quaardvark: There's a bunch of other unrelated files changed for some reason. @henriquecfreitas can we revert all changes not related to `Testimonials v2` item?

src/components/organisms/Testimonials/carousel/index.js
pkrawc: This comment refers to the entire file, not the top import line haha.

src/components/partials/SliceZone/components/TestimonialsSlice/index.js
quaardvark: When we talked with Matt I got the impression of the opposite, - we keep them as sibling slices

src/components/partials/SliceZone/components/TestimonialsSlice/index.js
quaardvark: I don't really the point of having `<Testimonials version={} />`, just pull the right component in the right slice

src/components/elements/Anchor/index.js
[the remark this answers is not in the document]
henriquecfreitas: If you feel that its important we can revert it, but they are only lint fixes applied by `eslint --fix`.

src/components/molecules/BoxedTitleSection/index.js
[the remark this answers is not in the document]
henriquecfreitas: Just pushed the refactor. I also think that its way better now, with flex

src/components/elements/Anchor/index.js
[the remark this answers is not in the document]
quaardvark: Please revert, thank you

quaardvark:
@henriquecfreitas Can you please cross-check the paddings/spacing on this PR with the Figma mockups? I see some inconsistencies, at least in `BoxedTitleSection`

src/components/organisms/Testimonials/cards/index.js
henriquecfreitas: @pkrawc I think those items can be a good discussion for us to have.
I think that maybe creating a `useTestimonialsCards` or `useCards` to handle the logic for `TestimonialsCards` would work better then creating a `useTestimonials`, since all the logic for `TestimonialsCarousel` is handled by `useCarousel`, and therefore we would have no use for a `useTestimonials` hook on the other implementation of the testimonials organism.
I would like to discuss the suggestion#2 also.

But I don't think that these discussions should be placed here, on this Pull Request, since they are more related to the Cards implementation than the Carousel (which is the subject of this PR).
Maybe we can create some Issues related to the TestimonialsCards and use its threads to discuss each one of the items you pointed here. How do you feel about that?

src/components/organisms/Testimonials/cards/index.js
henriquecfreitas: Same as https://github.com/rentalutions/marketing/pull/44#discussion_r582908827

src/components/organisms/Testimonials/carousel/testimonials-carousel-item.js
henriquecfreitas: @pkrawc I just applied this approach locally and as far as I understood of it the results are the same as they was on the direct definition using a `switch` or the `if`'s  approach I pushed. Right?

When I first saw your suggestion, I thought that the values on `opacity` and `scale` would be handled to provide some kind of transition/animation. Was that your intention? Because maybe I did something wrong on the usage of this code, and could take full advantage of the approach you suggested.

But if the results are meant to be the same, I would prefer to keep using something like the approach I pushed, because it seems more directly understandable to me, since if a third developer reads this code, it won't require him to know what `useSpring` and `useTransform` hooks are receiving and returning.

henriquecfreitas:
> @henriquecfreitas Can you please cross-check the paddings/spacing on this PR with the Figma mockups? I see some inconsistencies, at least in `BoxedTitleSection`

Lots of issues fixed on the last commit!
Can you please review it again?

Thank you for noticing this!

src/components/organisms/Testimonials/carousel/testimonials-carousel-item.js
pkrawc: I was thinking that it would animate the values yeah. But I see now that we're not really moving each item, we're just reassigning the image src to the item next in line.

pkrawc (approved):
I'm approving this PR based on conversations in slack.

src/components/organisms/Testimonials/cards/index.js
henriquecfreitas: To be addressed by #58

src/components/organisms/Testimonials/carousel/testimonials-carousel-item.js
henriquecfreitas: To be addressed by #58

src/components/partials/SliceZone/components/TestimonialsSlice/index.js
henriquecfreitas: To be addressed by #58

quaardvark (approved):

PR #48, opened 2021-02-02
Adding all color options to selects

henriquecfreitas:


henriquecfreitas:
@quaardvark @pkrawc I spent some time trying to find a way to allow the user to select any custom hex RGB color with the Prismic color element.
Would be great if we could enable/disable a field based on the selection of a previous field, so we would allow the Color element to be used only when the user selects "Custom" as an option from the Select element.
Another way would be just using the Color element, but have our system colors pre-defined on it.
None of those approaches is possible on Prismic right now, but I'm open to suggestions if you guys have any.

I didn't created the Color element bellow the Select element (with a "Custom" option that would make the code consider the value from the Color element) because I thought that this would be more confusing than useful for the user.

quaardvark (commented):
Are `blue_700` and `blue_900` added? I can't see them for some reason

I think that we might opt for automatic text color selection, based on background, but this is out of scope of this ask

henriquecfreitas:
> Are `blue_700` and `blue_900` added? I can't see them for some reason
> 
> I think that we might opt for automatic text color selection, based on background, but this is out of scope of this ask

I didn't saw any of those two in previous color select fields, so I assumed that they were not system colors.
But I can add it, sure

henriquecfreitas:
> I think that we might opt for automatic text color selection, based on background, but this is out of scope of this ask

I like the idea. Maybe instead of presenting bg and text colors options, we present color schemas options

PR #49, opened 2021-02-04
Fixing missing thumbnail on Safari

henriquecfreitas:


quaardvark (changes requested):
Can't we just use the timestamp hack for all devices? That way we wont need device detection

In any case, if we HAVE to use device detector, I would very much prefer a dependency instead of custom hook.

henriquecfreitas:
> Can't we just use the timestamp hack for all devices? That way we wont need device detection

Yes, we can.
But this timestamp hack has a side effect: it will override the defined thumbnail with the first frame.
I think that we should be using it only in necessary cases.


> In any case, if we HAVE to use device detector, I would very much prefer a dependency instead of custom hook.

I agree. I also though about using it and even searched for some options that works well on react projects.

quaardvark (approved):

PR #53, opened 2021-02-17
Implementing RichText embed content on FAQ Slice

henriquecfreitas:


src/components/partials/SliceZone/components/Embed/index.js
```diff
@@ -1,17 +1,6 @@
...
-    <ResponsiveEmbed
-      aspect={{ width: embed.width, height: embed.height }}
-      {...props}
-    >
-      <RichText render={[{ type: "embed", oembed: embed }]} />
-    </ResponsiveEmbed>
-  ) : (
-    "Embed provider not supported"
-  )
+  return <RichText {...props} render={[{ type: "embed", oembed: embed }]} />
```
quaardvark, line 5: This should happen vice versa. Rich Text htmlSerializer should render `<Embed />`. We are trying to keep components reasonably vendor-agnostic, which is why we never use RichText in them. RichText is CMS specific code and we should be able to swap it out.

src/components/partials/SliceZone/components/FrequentlyAskedQuestionsSlice/index.js
```diff
@@ -10,7 +11,17 @@ export default function FrequentlyAskedQuestionsSlice({ slice }) {
 
   const questions = slice.items.map(({ question, answer }) => ({
     question,
-    answer: <RichText render={answer} />,
+    answer: (
+      <RichText
+        render={styleEmbed(answer, {
```
quaardvark, line 16: This is not a good API. `RichText` props are forwarded to `htmlSerailizer` and you can use them cleaner there, where the actual rendering happens.

quaardvark (changes requested):

src/components/partials/SliceZone/components/Embed/index.js
henriquecfreitas: Yes, I agree, actually this was the first think I thought to do.
But then we would have to handle it on `RichText` to avoid them to be calling each other recursively. Since it would be a bit messing, I preferred to just handle everything on  `RichText` and make `Embed` call it.
But I agree with you, so I'll think of good approach to do it on `Embed` and come back with a solution here

henriquecfreitas: Do you agree on this premises?

1. Embed may be called directly from other components, by fields that aren't `RichText`
2. Embed may be called inside `RichText`, from `htmlSerializer`
3. Embed should not call `RichText` anymore, and the `embed` logic that was previously handled by Primic's `RichText` should be done on our `Embed` component now

src/components/partials/SliceZone/components/FrequentlyAskedQuestionsSlice/index.js
henriquecfreitas: But this style props were meant to be applied only on elements of the type `Embed` inside the RichText.
In this case, I applied it on the video to make it smaller on large screens, since the FAQ `Answer` component should take 100% of the container's width. My intention here was to apply it to the `Embed`, but keeping the `Header` with its original style.
If I use the `RichText` props, they'll be applied to all elements of the `RichText`, including the `Header`. Also, I think that these props wouldn't have priority over the ones that are being handled by `ResponsiveEmbed`.

quaardvark: I suggest that we do something like this:
```
<RichText embedSx={{}} />
```

And then in `htmlSerializer`:
```
<Embed
  {...props}
  sx={{
    ...props.sx,
    ...props.embedSx,
  }}
>
```

That way the whole logic with going through the elements and injecting it there can be omitted

henriquecfreitas: Great approach, we can go with it.

src/components/partials/SliceZone/components/Embed/index.js
henriquecfreitas: But I don't think that `<Embed />` should be Prismic-specific anymore, considering these proposed changes...
Relying on these premises we may create a reusable and independent component that shouldn't  be affected if we replace Prismic by anything else.

How do you feel about this?

quaardvark: It's purpose is to validate that `embed` provider is supported by slices. The actual element-like component is `ResponsiveEmbed`

src/components/elements/ResponsiveEmbed/index.js
```diff
@@ -25,10 +25,11 @@ const Wrapper = ({ sx, children }) => (
 export default function ResponsiveEmbed({
   aspect: { width = 100, height = 100 } = {},
   sx,
+  embedSx,
   children,
 }) {
   return (
-    <Wrapper sx={{ pb: `${(height / width) * 100}%`, ...sx }}>
+    <Wrapper sx={{ pb: `${(height / width) * 100}%`, ...sx, ...embedSx }}>
```
quaardvark, line 32: I don't think `embedSx` should be used here? Is it only for a story? Why not just use `sx`?

src/components/organisms/FAQ/index.js
```diff
@@ -93,7 +93,19 @@ function FAQ({
...
+                          typeof Answer === "function" ? <Answer /> : Answer,
+                          {
+                            embedSx: {
+                              maxWidth: "65rem",
+                              height: "unset",
+                              margin: "auto",
+                              aspectRatio: "16/9",
+                              paddingBottom: 0,
+                              marginTop: 1,
+                            },
```
quaardvark, line 106: This feels out of place here. It takes any elements as an answer and shouldn't really be concerned with embeds. Maybe move it into `<RichText />` in `FaqSlice` and `embedSx` should also be contained to `htmlSerializer`, IMO.

quaardvark (changes requested):

quaardvark (approved):
:tada:

PR #56, opened 2021-02-23
Add data capture capability to the application

quaardvark:


pkrawc:
Hey, looking at this now. Sorry it's been like 12 days.

1. Do we actually need apollo? Mostly okay with it even if not but just want to make sure I'm not missing something.
2. What's the goal of this PR? I don't see much to do with forms here. Is it just making sure that we can share cookies across avail.co and info.avail.co?

quaardvark:
@pkrawc 

1. We can make do with context, - would give smaller package. I just thought we would eventually have Apollo anyway, so why not utilize it's capability to be a store.
2. Forms are no longer the hot topic, this is more about data capture with the right cookies/IDs right now

pkrawc (approved):
I'm fine with the graphql, even if it is just a wrapper around a rest endpoint for now.

I can't see anything that I would change necessarily, I wonder at the need for both a jsconfig and the babel module resolution. If there's a way to declare the resolution patterns in one place instead of two that would be my preference.

quaardvark:
@pkrawc I've added the actual Segment submission part, sorry about that.

About the rest endpoint - it's a very small thing here, it's only purpose is to validate/decode the shared JWT payload and return it. I thought setting up Apollo Server would be overkill for that. I'm mostly interested in Apollo cache as a store here. Also because the moment Prismic GraphQL is out of beta I would love to switch the pages to Apollo as well.

henriquecfreitas (approved):
Approved!
I would only rename `pages/api/analytics.js` to something more related to the JWT validation, since it seems to be what the function is currently doing. Also because it would be better to have a different file name to avoid confusion with `utils/analytics.js` (for humans and IDE's)

PR #59, opened 2021-03-01
Replacing image references to media references

henriquecfreitas:


src/types/info.json
```diff
@@ -481,12 +481,12 @@
                   "label" : "Embed"
                 }
               },
-              "imagePosition" : {
+              "mediaPosition" : {
```
quaardvark, line 484: We shouldn't rename fields in Prismic, the original value would be lost in this case. Let's rename only placeholders / labels/

quaardvark (changes requested):

quaardvark (approved):

PR #65, opened 2021-03-11
Temporary solution to remove the a:link style tag

henriquecfreitas:
Link to the issue: https://avail-oc.atlassian.net/servicedesk/customer/portal/4/PS-1567
Reproducible at: https://info.avail.co/info/resources-tenants-renting-information-hub

PR Explanation: **This is not a definitive solution**, but since this bug is currently affecting production mode of the CMS app, I propose we merge and deploy this temporary workaround that will have no impact on the application. This should give us some more time to investigate who is injecting the unwanted style tag on the document head.

I would need access to Vercel deployment information in order to try to check how and what third part libs are being injected on the application and why I can see them on my local dev environment.

### DIFF
--- a/pages/_document.js
+++ b/pages/_document.js
@@ -90,6 +90,26 @@ export default class Document extends NextDoc {
           <Main />
           <NextScript />
         </body>
+        {/**
+            The script bellow was inserted as an urgent and **temporary** fix
+            to solve the bug 'Prismic Primary Button Variation only displays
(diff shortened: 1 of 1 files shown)


pkrawc:
I like where our heads at but I would rather explore Digioh / Tag Manager to determine what is setting the style tag instead of adding something that opens us up for SSX attacks in the future.

henriquecfreitas:
Issue solved without changes on the code

PR #67, opened 2021-03-15
Add an option to set page background color

henriquecfreitas:


quaardvark:
@henriquecfreitas @pkrawc I'm wondering if setting the background on wrapper is good enough, - I would prefer it on the body to be honest

pkrawc:
I would also prefer it on the body.

@henriquecfreitas one can create a global style tag anywhere in the app with `createGlobalStyle`

I would suggest the following.

```jsx
const BodyStyles = createGlobalStyle(
  ({ bg }) => css`
    body {
      background: ${bg};
    }
  `
)

function InfoPage() {
  return (
    <Fragment>
      <BodyStyles bg={background} />
      <Box /> // Page Stuff
    </Fragment>
  )
}
```

You cool with that @quaardvark ?

pkrawc:
Almost 100% that the global style tag will be removed from the DOM when the component is unmounted. But that's something we should probably check for.

pkrawc:
You'll also probably want to add a default option to that prop (eg. `bg = "#f3f3f3"`)

quaardvark (approved):

PR #68, opened 2021-03-16
Creating Avail/RDC header variant

henriquecfreitas:
New Avail/RDC navbar:
Screen Shot 2021-03-17 at 3 44 51 PM
Screen Shot 2021-03-17 at 3 44 43 PM

Old Avail navbar (still working as normal):
Screen Shot 2021-03-17 at 3 45 21 PM
Screen Shot 2021-03-17 at 3 45 15 PM


src/types/info.json
```diff
@@ -24,6 +32,50 @@
...
+      "config" : {
+        "label" : "Primary Button hash (jump link)",
+        "placeholder" : "A hash matching the hash in slice you want to link to"
+      },
+      "type" : "Text"
+    },
+    "secondaryButtonText" : {
+      "type" : "Text",
+      "config" : {
+        "label" : "Primary Button text",
```
quaardvark, line 60: Should be "Secondary Button Text" probably

src/types/info.json
```diff
@@ -24,6 +32,50 @@
...
+      },
+      "type" : "Link"
+    },
+    "secondaryButtonHash" : {
+      "config" : {
+        "label" : "Secondary Button hash (jump link)",
+        "placeholder" : "A hash matching the hash in slice you want to link to"
+      },
+      "type" : "Text"
+    },
```
quaardvark, line 78: Are there plans to discontinue NavBar v1? If not, I would prefer to have consistent API, instead of supporting two versions

src/components/organisms/NavBar/AvailNavBar/index.js
```diff
@@ -0,0 +1,133 @@
+import React from "react"
```
quaardvark, line 1: This component errors out with 
Screenshot_20210318_173340

pages/info/[uid].js
```diff
@@ -116,10 +122,17 @@ const Page = ({ data, uid }) => {
...
         borderBottom={navBarSticky ? `1px solid ${colors.ui_500}` : "none"}
         containerWidth={CONTAINER_WIDTHS}
+        type={navBarType}
+        sticky={navBarSticky}
+        primaryButtonText={primaryButtonText}
+        primaryButtonLink={primaryButtonLink}
+        primaryButtonHash={primaryButtonHash}
+        secondaryButtonText={secondaryButtonText}
+        secondaryButtonLink={secondaryButtonLink}
+        secondaryButtonHash={secondaryButtonHash}
```
Comment (quaardvark) [inline pages/info/[uid].js:134]: Can we squash these into objects?

src/components/organisms/NavBar/index.js
```diff
@@ -38,22 +23,60 @@ const NavBarButton = styled(Button)`
...
-  sticky,
   animationPreset = "fadeIn",
+  type = "Avail",
+  sticky,
+  primaryButtonText: pbText,
+  primaryButtonLink: pbLink,
+  primaryButtonHash: pbHash,
+  secondaryButtonText: sbText,
+  secondaryButtonLink: sbLink,
+  secondaryButtonHash: sbHash,
```
quaardvark, line 35: Squishy squashy

quaardvark (commented):
@pkrawc @henriquecfreitas 
Is this misalignment deliberate or accidental?
Screenshot_20210318_174910

I don't have a lot of objections. Thanks for making the changes backwards-compatible, but I feel that we can actually ditch the whole "primary: true" option on links and just use the static fields for these. Otherwise the API is too confusing

pkrawc:
> Is this misalignment deliberate or accidental?
> Screenshot_20210318_174910

What is generating those tracks @quaardvark? Probably yes intended, it looks like an accuratete copy of the codesandbox I proposed.

https://codesandbox.io/s/intelligent-tereshkova-nujdh?file=/src/header.jsx

src/components/organisms/NavBar/AvailRdcNavBar/index.js
```diff
@@ -0,0 +1,206 @@
...
+        <Col
+          display="flex"
+          span={[5, 4, 3]}
+          padding="2rem 4rem 2rem 2rem"
+          bg="blue_900"
+          justifyContent="flex-end"
+          alignItems="center"
+          sx={{
+            clipPath: `polygon(0 0, calc(100% - 2rem) 0, 100% 100%, 0% 100%)`,
+          }}
```
pkrawc, line 69: Why did we choose to use system props instead of the sx object for styling here?

src/components/organisms/NavBar/AvailRdcNavBar/index.js
```diff
@@ -0,0 +1,206 @@
...
+                >
+                  <Button
+                    variant="primary"
+                    display="block"
+                    flex="0 0 auto"
+                    {...primaryLinkProps}
+                  />
+                </Box>
+              )}
+              <Menu>
```
pkrawc, line 148: This menu component would do well pulled out of the parent I think

src/types/info.json
pkrawc: I think this is never going to fully replace all navbars, and I would agree with the consistent API comment.

quaardvark:
@pkrawc These are just guides I've put in GIMP, no issues on my part if intended

quaardvark:
@henriquecfreitas also one more thing - the links should support the ID field, just like other standalone buttons

pkrawc:
Gotcha. Initially, they looked like the grid tracks you get when inspecting a page in chrome or firefox. But yeah layout-wise, seems right to me.

@quaardvark

src/components/organisms/NavBar/AvailRdcNavBar/index.js
```diff
@@ -0,0 +1,206 @@
...
+                flexDirection: "row",
+                alignItems: "center",
+                "& > *:not(:first-child)": {
+                  marginLeft: "2rem",
+                },
+              }}
+            >
+              {secondaryLink && (
+                <Box
+                  as="motion.li"
```
quaardvark, line 123: This cannot be a string, - it will render <motion.li> tag instead of `motion.li` component, - should be `as={motion.li}`

henriquecfreitas:
> @pkrawc @henriquecfreitas
> Is this misalignment deliberate or accidental?
> Screenshot_20210318_174910

Not intentional. First I tried to keep the same width from the old header, but it was bigger than the mock from code sandbox.
Then I removed the `maxWidth` (that was limited to the container width) and it shrunk  to the right size.
We can place the `maxWidth` back to compare results and choose which one is better.

> I don't have a lot of objections. Thanks for making the changes backwards-compatible, but I feel that we can actually ditch the whole "primary: true" option on links and just use the static fields for these. Otherwise the API is too confusing

Yeah, I would also prefer we use only the static fields, but two things to keep in mind:
1. We'll loose the primary buttons that are current defined by the primary boolean, unless we can migrate that data some way on Prismic.
2. How will we handle the static secondary link fields on the old navbar? There will be no difference from it to the repeatable zone links (but the position?)

src/types/info.json
henriquecfreitas: Do we want to use the co branded NavBar on all pages?
I'm not sure, but I think I record Matt saying something about using it on specific pages for now.

henriquecfreitas: But I agree @quaardvark, I'm also not 100% comfortable in having so many fields and mainly with having fields with different proposes based on the NavBat Type.

Does it make sense to think about creating a new content type?

src/components/organisms/NavBar/AvailRdcNavBar/index.js
henriquecfreitas: Yes it will, good call

src/components/organisms/NavBar/AvailRdcNavBar/index.js
henriquecfreitas: I though that declaring them as element props would be better, but no specific reason on that.
I'll place them on `sx`, for this component and others

quaardvark:
@henriquecfreitas For the layout Patrick says it's OKay.

For the NavBar API - I will just convert old NavBars to new API manually on release, - I don't think we have "secondary" links currently on production pages, so it's mostly about copying "primary" button to new API

src/components/organisms/NavBar/AvailNavBar/index.js
henriquecfreitas: I think this was caused by two different links having the same `href`.
This should not happen on production, but maybe we can change the `href` key to a compound key of `text-href`?

src/components/organisms/NavBar/index.js
```diff
@@ -38,22 +23,53 @@ const NavBarButton = styled(Button)`
...
-     * the rest would be considered default links */
-    const primaryLinkIndex = _defaultLinks.findIndex(({ primary }) => primary)
-    const [_primaryLink] =
-      primaryLinkIndex !== -1 ? _defaultLinks.splice(primaryLinkIndex, 1) : []
+    const _menuEntries = _defaultLinks.map((a) => a)
+    const _secondaryLink = secondaryButton && createLink(secondaryButton)
+
+    if (_secondaryLink) {
+      _defaultLinks.push(_secondaryLink)
+    }
```
quaardvark, line 72: It should be possible to ditch and/or simplify this logic with cutting over from `push: true, primary:true` to explicitly defined primary + secondary buttons

quaardvark (changes requested):

src/components/organisms/NavBar/AvailNavBar/index.js
```diff
@@ -0,0 +1,133 @@
...
+
+function AvailNavBar({
+  background = "ui_100",
+  defaultLinks = [],
+  primaryLink,
+  pushIndex,
+  containerWidth = "96rem",
+  sticky,
+  animationPreset = "fadeIn",
+  ...props
```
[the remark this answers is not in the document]
henriquecfreitas: No.
On the new navbar we have only one secondary button, and the repeatable links are rendered as menu items.
On the old navbar both the static secondary link and all the repeatable links are rendered as secondary buttons, with no differences between them. So we only have to push the secondary link to the `defaultLinks` array.

src/components/organisms/NavBar/AvailRdcNavBar/index.js
```diff
@@ -44,12 +44,13 @@ function AvailRdcNavBar({
...
   return (
-    <Flex
+    <Box
       as={motion.header}
       {...animation?.container}
       ref={intersectionTarget}
+      bg={background}
       sx={{
-        background,
+        display: "flex",
```
pkrawc: Interesting that background isn't doing what you'd expect.

bg in the sx object should work though.

pkrawc (changes requested):

quaardvark (approved):

PR #74, opened 2021-03-31
Adding Source and Medium UTMs URL vars to info pages

henriquecfreitas:
Resolves: https://airtable.com/tbl4CgbPGPmsSc1pl/viw6xcE860VULXMPz/rece4GbMCfXIzCJQc?blocks=hide


quaardvark (changes requested):
It seems like this is not enough to make it work

PR #78, opened 2021-04-01
Add page-specific params to the analytics traits

quaardvark:


quaardvark:
@pkrawc @henriquecfreitas I'm going to fast-track this one, but it's worth a look. I used a pretty amazing thing here - https://www.apollographql.com/docs/react/local-state/reactive-variables/

PR #87, opened 2021-04-14
Creating Wiki

henriquecfreitas:


pkrawc:
@henriquecfreitas I just updated the case for all pages. Let's stick with all lowercase from here on out, and proposed changes to the prismic custom types docs.

I don't think it covers all the use cases of custom types, we're thinking of custom types as only page types, is that all there is? Does SEO and Metadata apply to non-page custom types?

pkrawc (commented):
I think some changes are still needed, this isn't priority though.

PR #91, opened 2021-04-26
Add GQL fetch optimization, expose UUID as global variable

quaardvark:
This achieves 2 things:

1) Exposes UUID as a global variable for Digioh tracking purposes
2) Adds a requests optimization mechanism for concurrent local resolvers

### DIFF
--- a/src/graphql/resolvers/client.js
+++ b/src/graphql/resolvers/client.js
@@ -1,0 +1,58 @@
+import objectHash from "object-hash"
+import { ApolloError, throwServerError } from "@apollo/client"
+
+const IN_PROGRESS_REQUESTS = {}
+
+function parseBody(input) {
+  if (/^\s*<!doctype/i.test(input)) {
+    return ""
+  }
+  try {
+    return JSON.parse(input)
+  } catch (e) {
+    return input
+  }
+}
+
+async function execRequest({ url, ...params }) {
+  try {
+    const response = await fetch(url, params)
+    const responseBody = parseBody(await response.text())
+    if (!response.ok) {
+      throwServerError(
+        response,
+        responseBody,
+        responseBody?.message || "Unknown request failure"
+      )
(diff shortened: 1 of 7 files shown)


henriquecfreitas (approved):

src/graphql/resolvers/client.js
```diff
@@ -0,0 +1,59 @@
...
+    return JSON.parse(input)
+  } catch (e) {
+    return input
+  }
+}
+
+async function execRequest({ url, ...params }) {
+  try {
+    const response = await fetch(url, params)
+    const responseBody = parseBody(await response.text())
```
pkrawc, line 21: Why not just `response.json()` here and forgo the need to parse the body?

src/graphql/resolvers/client.js
```diff
@@ -0,0 +1,59 @@
...
+  const hash = objectHash(params)
+  let request = inProgressRequestsCache.get(hash)
+  if (!request) {
+    request = execRequest(params).finally(() =>
+      inProgressRequestsCache.del(hash)
+    )
+    inProgressRequestsCache.set(hash, request)
+  }
+  return request
+}
```
pkrawc, line 59: Is this function just to make sure that we're always including credentials in each request?

pkrawc (commented):
Not positive on a couple of pieces but likely that's because I lack context.

src/graphql/resolvers/client.js
quaardvark: If you'll get a 404 (or any other) page response as HTML it will throw the stack on parse instead of letting you process the 404

src/graphql/resolvers/client.js
quaardvark: You are kidding, right? It's there to squash concurrent requests with the same parameters to one

PR #93, opened 2021-05-12
Feature/blockquote component

henriquecfreitas:
Screen Shot 2021-05-12 at 3 18 01 PM
Screen Shot 2021-05-12 at 3 17 30 PM
Screen Shot 2021-05-12 at 3 17 45 PM


src/components/organisms/Blockquote/index.js
```diff
@@ -0,0 +1,48 @@
...
+            position: "absolute",
+            marginLeft: "2rem",
+            width: "7rem",
+          }}
+        >
+          <DoubleQuotes fill={quoteColor} />
+        </Box>
+        {content && (
+          <Box as={motion.aside} {...animation?.item} sx={{ textAlign }}>
+            {cloneElement(content, {
```
pkrawc, line 34: This is assuming the content will always be an element that consumes the sx prop yeah? Probably fine but just want to make sure I'm understanding correctly.

pkrawc (approved):
As long as there are not issues with the one comment I made this looks like great work. I'm not sure if there was a right-aligned option proposed in the mocks, but if not I like the initiative.

src/components/organisms/Blockquote/index.js
henriquecfreitas: Yes, the content is supposed to be a `RichText` element.
Titles and other text content props are being handled this way by other organisms, like `ButtonCTA` and `Hero`, so I just followed this line.
But I can adjust it so it can also work with a string content, what do you think?

quaardvark:
@pkrawc @henriquecfreitas is this supposed to be non-Prismic component only? Are we planning to also implement a slice for this?

henriquecfreitas:
> @pkrawc @henriquecfreitas is this supposed to be non-Prismic component only? Are we planning to also implement a slice for this?

My head was so immersed in recreating pages directly in Next.JS that I totally forgot to create the `SliceZone` component.
I'm sorry for that, and thank you a lot for noticing it!

src/components/organisms/Blockquote/index.js
pkrawc: Looks like the only element `RichText` could return that doesn't consume the sx prop is an anchor tag, that's probably fine. No change needed here.

src/types/info.json
```diff
@@ -154,6 +155,57 @@
...
+                  "default_value" : "none",
+                  "label" : "Background color"
+                }
+              },
+              "quoteColor" : {
+                "type" : "Select",
+                "config" : {
+                  "options" : [ "green_100", "green_300", "green_500", "green_700", "green_900", "ui_100", "ui_300", "ui_500", "ui_700", "ui_900", "blue_100", "blue_300", "blue_500", "blue_700", "blue_900" ],
+                  "default_value" : "green_100",
+                  "label" : "Quote color"
```
quaardvark, line 188: Let's be more explicit and say "Quote symbol color"

src/components/partials/SliceZone/components/BlockquoteSlice/index.js
```diff
@@ -0,0 +1,23 @@
...
+const BlockquoteSlice = ({ slice }) => {
+  const {
+    primary: { content, background, quoteColor, textColor, textAlign = "left" },
+  } = slice
+
+  return (
+    <Blockquote
+      containerWidth={CONTAINER_WIDTHS}
+      content={content && <RichText render={content} />}
+      bg={background === "none" ? undefined : background}
```
quaardvark, line 15: Why not use `transparent`? What does `none` means?

quaardvark (changes requested):
@henriquecfreitas @pkrawc is this the design this PR is implementing? https://www.figma.com/file/tyf4c9SVsRKSqxTEOyTDkb/Marketing-Organisms?node-id=1295%3A644

I have couple of questions if yes:
1) For `Variant B` the quote symbol positioning seems to be off
2) In Prismic it is currently implemented as a heading and H1 by default. I'm not sure if this is semantically correct. Semantically, heading is followed by either plain text or lower-level headings. I think maybe we should leave H1...5 off here, threat is a plain text and customize the font size purely via labels with the default font size being driven by the mock up. Thoughts?
3) Right aligned text looks weirder on desktop. Maybe the quote sign should follow the text?
Screenshot_20210517_153316

4) Are the other variants (A and C) planned for the future?

henriquecfreitas:
> @henriquecfreitas @pkrawc is this the design this PR is implementing? https://www.figma.com/file/tyf4c9SVsRKSqxTEOyTDkb/Marketing-Organisms?node-id=1295%3A644
> 
> I have couple of questions if yes:
> 
> 1. I see that the PR is opened against `pages/main` - why don't we open it against `staging`, release it as a slice and merge it into `pages/main` after release?
> 2. For `Variant B` the quote symbol positioning seems to be off
> 3. In Prismic it is currently implemented as a heading and H1 by default. I'm not sure if this is semantically correct. Semantically, heading is followed by either plain text or lower-level headings. I think maybe we should leave H1...5 off here, threat is a plain text and customize the font size purely via labels with the default font size being driven by the mock up. Thoughts?
> 4. Right aligned text looks weirder on desktop. Maybe the quote sign should follow the text?
>    Screenshot_20210517_153316
> 5. Quote sign is not animated by the in-view animation. I think it would look better if it's also faded in with the text
> 6. Are the other variants (A and C) planned for the future?

1. I agree. We'll need to merge `staging` into `pages/main` anyway to get the `Slider` component. This one goes for the same rule. (and this also applies for the `Article List` component I'll be opening a PR soon)
2. You mean `Variant B` on the Figma or on the test page? If you're talking about the right aligned variant, I'll answer it on item number 4.
3. I though that handling it as a `RichText` would be better, in order to give the user more options on how to style each quotation. But ok, we can also go with plain text.
4. Right aligned text was not previously contemplated on the Figma design (I took the liberty to implement it, based on the fact that other components handles orientation/alignment properties). Therefore I can't say if leaving the quote on the left was wrong or right. I felt it would be better this way, because the user will be reading the text from left to right regardless of its alignment (at least until we start writing content in Arabic, Japanese or other RTL languages).
Also, it will look less weird with real content, since those quotations should probably have more than 4 words.
Anyway, I agree that the quote sign looks weird and could be better displayed on the case of that test page, and we might need some help from @pkrawc or Vini, on that. If you guys prefer, I can just remove the right align variation for now.
5. This was intentional. I tested both possibilities and IMHO it looks better this way, the quote sign is displayed as like being a part of the background and the text fades into it. We should open this topic for discussion but I really liked this way 😁
6. I'm not sure if I'm aware of those variants. Where did you saw them?

src/components/partials/SliceZone/components/BlockquoteSlice/index.js
henriquecfreitas: It means no background 😁
But I'll change it to `transparent`, I see we already have a precedent for it on `How It Works` background options.

quaardvark:
> 1. I agree. We'll need to merge `staging` into `pages/main` anyway to get the `Slider` component. This one goes for the same rule. (and this also applies for the `Article List` component I'll be opening a PR soon)
> 2. You mean `Variant B` on the Figma or on the test page? If you're talking about the right aligned variant, I'll answer it on item number 4.
> 3. I though that handling it as a `RichText` would be better, in order to give the user more options on how to style each quotation. But ok, we can also go with plain text.
> 4. Right aligned text was not previously contemplated on the Figma design (I took the liberty to implement it, based on the fact that other components handles orientation/alignment properties). Therefore I can't say if leaving the quote on the left was wrong or right. I felt it would be better this way, because the user will be reading the text from left to right regardless of its alignment (at least until we start writing content in Arabic, Japanese or other RTL languages).
>    Also, it will look less weird with real content, since those quotations should probably have more than 4 words.
>    Anyway, I agree that the quote sign looks weird and could be better displayed on the case of that test page, and we might need some help from @pkrawc or Vini, on that. If you guys prefer, I can just remove the right align variation for now.
> 5. This was intentional. I tested both possibilities and IMHO it looks better this way, the quote sign is displayed as like being a part of the background and the text fades into it. We should open this topic for discussion but I really liked this way grin
> 6. I'm not sure if I'm aware of those variants. Where did you saw them?

1. Great!
2. Yes, the position of the quote sign looks different relative to the text. I'm talking about the left aligned text variant
3. I mean "plain text" in a semantic sense, from the POV of content parser. It should be a RichText, but should not be a header (at least by default). Let's make it multi-paragraph `p` and add all the font-size labels to it.
4. I'm fine shipping it like this, but design team input would be valuable here.
5. Again, if the design team is fine I'm also fine, but it looks _super_ weird to me, because with the normal animations the page is blank and the component fades in as a whole
6. I linked the Figma file above, - it has variants A and C. Variant C looks a little bit weird, but variant A (with the vertical line) is what the current home page uses, - did we decide to go with different version for Home blockquote?

henriquecfreitas:
@quaardvark I just saw the other variants you're talking about. They were on the bottom of the Figma and I haven't noticed them until now. They are tagged as `Studies`, so I guess that the top frame (the one that only presents variant B) is the one meant to be implemented. I'm I right, @pkrawc ?

pkrawc:
Those other frames were drafts of the final, not to be implemented. I've hidden them for now, good question though.

pkrawc:
@henriquecfreitas I've made updates to the figma file to illustrate the alignment of the quote icon for the right aligned text.

PR #95, opened 2021-05-14
Email Capture features: Email validation, event tracking and opt in

quaardvark:


pages/info/[uid].js
```diff
@@ -123,6 +121,11 @@ const Page = ({ data, uid }) => {
       id: buttonId,
     })
   )
+
+  useEffect(() => {
+    analyticsVar({ ...analyticsVar(), ...urlResolverParams })
+  }, [urlResolverParams])
+
```
Comment (quaardvark) [inline pages/info/[uid].js:128]: This is a bugfix, apparently it's using some sort of state underneath and calling `setState` in the render flow is a no-no

src/components/molecules/EmailCaptureInput/index.js
```diff
@@ -3,6 +3,15 @@ import Input from "@rent_avail/input"
 import { Box } from "@rent_avail/layout"
 import Button from "components/elements/Button"
 import { analyzeColor } from "utils/color-scheme"
+import styled from "styled-components"
+import { Checkbox as AvailCheckbox } from "@rent_avail/controls"
+import { sx } from "@rent_avail/base"
+
+const Checkbox = styled(AvailCheckbox)`
+  ${sx}
+`
```
quaardvark, line 12: Adds missing `sx` to `Checkbox`

henriquecfreitas: Since we only have a single `sx` value being applied to `Checkbox`, I would just directly create a `StyledCheckbox` with the style values here.
Even better than that would be adding this `sx` handler to the `Checkbox` component on `@rent_avail/controls`

src/components/organisms/EmailCapture/index.js
```diff
@@ -31,16 +25,7 @@ function EmailCapture({
...
-          inputLabel={inputLabel}
-          inputLabelId={inputLabelId}
-          buttonText={buttonText}
-          buttonUrl={buttonUrl}
-          queryParamName={queryParamName}
-          analyticsParamName={analyticsParamName}
-          background={containerBg}
-          onSubmit={onSubmit}
-        />
+        <EmailCaptureInput background={containerBg} {...props} />
```
henriquecfreitas, line 28: My first feeling when passing props to a custom component is that any non-handled prop will be:
a) ignored
or
b) passed to the top level element (in this case, the `Box` wrapper)

Based on that, I would keep those attributes explicitly defined on `EmailCapture` props, and spread the other props to the `Box`

src/types/info.json
```diff
@@ -270,6 +270,23 @@
                   "placeholder" : "E.g. - https://www.avail.co/users/new"
                 },
                 "type" : "Text"
+              },
+              "optInContext" : {
+                "config" : {
+                  "label" : "Opt In Context (required to show opt in)",
+                  "placeholder" : "Email Capture Opt In, - e.g. The Landing"
+                },
+                "type" : "Text"
```
henriquecfreitas, line 279:
```suggestion
                "type" : "Text",
                "config" : {
                  "label" : "Opt In Context (required to show opt in)",
                  "placeholder" : "Email Capture Opt In, - e.g. The Landing"
                }
```

src/types/info.json
```diff
@@ -592,6 +609,23 @@
                   "placeholder" : "E.g. - https://www.avail.co/users/new"
                 },
                 "type" : "Text"
+              },
+              "emailCaptureOptInContext" : {
+                "config" : {
+                  "label" : "Opt In Context (required to show opt in)",
+                  "placeholder" : "Email Capture Opt In, - e.g. The Landing"
+                },
+                "type" : "Text"
```
henriquecfreitas, line 618:
```suggestion
                "type" : "Text",
                "config" : {
                  "label" : "Opt In Context (required to show opt in)",
                  "placeholder" : "Email Capture Opt In, - e.g. The Landing"
                }
```

src/utils/analytics.js
```diff
@@ -79,6 +79,30 @@ export function useAnalytics(params) {
           }
         })
       }
+      console.warn("Identify call not performed due to missing dependencies")
+      return Promise.resolve()
```
henriquecfreitas, line 83: Shouldn't this be a rejected promise?

src/utils/analytics.js
```diff
@@ -79,6 +79,30 @@ export function useAnalytics(params) {
...
+        const segment = segmentRef.current
+        const {
+          analytics: { uuid },
+        } = data
+
+        return new Promise((resolve, reject) => {
+          try {
+            segment.track(eventName, properties, { userId: uuid }, resolve)
+          } catch (e) {
+            console.error(e)
```
henriquecfreitas, line 100: Is this error log necessary?
Since we're already return the error on the reject, the caller will be able to handle it and log if necessary.
(also applies to corresponded code on `identify` useCallback)

src/utils/analytics.js
```diff
@@ -92,5 +116,28 @@ export function useAnalytics(params) {
...
+        Email: email,
+        "Email Capture Opt In": optInContext || undefined,
+        "Email Capture Opt In Checked": optInContext ? optIn : undefined,
+      }
+      await identify(traits)
+      await track("Marketing Site Form Submission", {
+        Type: "Email Capture",
+        ...defaultProperties,
+        ...traits,
+      })
```
henriquecfreitas, line 137: Do we want to wait for `identify` before calling `track`? Or abort the second call if the first one fails?
If not, we could make those calls independently:

```suggestion
      await Promise.all([
        identify(traits),
        track("Marketing Site Form Submission", {
          Type: "Email Capture",
          ...defaultProperties,
          ...traits,
        })
      ])
```

src/components/molecules/EmailCaptureInput/index.js
quaardvark: Thanks for suggestion, the last part is currently being implemented, but since we stick to `sx` the idea is to treat the `Checkbox` as if it had `sx`. When the `controls` package will get `sx` all is to be done here is to remove this reassignment. I don't see any value in singling out this checkbox to a separate `Styled...` reference, since it's not reused.

src/types/info.json
quaardvark: This is how Prismic formats it, there's no value in this suggestion since it will be overwritten anyway

src/types/info.json
quaardvark: Same a above

src/utils/analytics.js
quaardvark: It will prevent the actual submission, which is not a good idea. Better to have an untracked submission than nothing at all.

src/utils/analytics.js
quaardvark: Error log here is just for awareness for developers. I don't expect it to happen in actual operation, so it wouldn't be handled in consumer components, since it has no business meaning. Ideally we should throw it to the Honeybadger, but we are yet to set it up.

src/utils/analytics.js
quaardvark: Yes, we want `identify` before `track`, because we want to make sure the user is identified before tracking events for him

src/components/organisms/EmailCapture/index.js
quaardvark: You are right, I must have done it on autopilot

henriquecfreitas: I saw you reverted that, but maybe forgot to add and pass the new `optInContext` and `optInCopy`props?

src/types/info.json
henriquecfreitas: Got it!
Feels strange having different order for the fields depending on its type

src/utils/analytics.js
henriquecfreitas: Got it!

src/components/organisms/EmailCapture/index.js
quaardvark: Oh yeah, thanks for catching that. Good eye!

henriquecfreitas (approved):

PR #98, opened 2021-05-20
Performance Improvements

quaardvark:
In this pull request:

1) Upgraded Next.js to 10.2+, which includes automatic Google Webfonts optimization
2) Deferred all analytics processing to remove blocking `noop.js` Avail cookie script
3) Removed Apollo to reduce bundle size
4) Disabled animations for NavBar and first Hero slices (gained 10-ish score in Lighthouse)
5) Switched all images to `next/image` with defined `width` and `height`
6) Some minor fixes/adjustments here and there


pkrawc:
I do see improvement from the 9 score average, but it seems we've only improved it to 20? Is that expected?

https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fmarketing-86cnl376q-rent-avail.vercel.app%2Finfo%2Fbigger-pockets-sponsorship

Are we thinking there will be greater improvements when we kill some of the 3rd party scripts in tag manager?

quaardvark:
> I do see improvement from the 9 score average, but it seems we've only improved it to 20? Is that expected?
> 
> https://developers.google.com/speed/pagespeed/insights/?url=https%3A%2F%2Fmarketing-86cnl376q-rent-avail.vercel.app%2Finfo%2Fbigger-pockets-sponsorship
> 
> Are we thinking there will be greater improvements when we kill some of the 3rd party scripts in tag manager?

I got 53 there, which is consistent with what I was seeing: Screenshot_20210521_174425 



With 3-rd party disabled I'm getting consistent 90-95+ with local Lighthouse 
Screenshot_20210521_175114



and ~75-ish on Insights
https://lighthouse-dot-webdotdevsite.appspot.com//lh/html?url=https%3A%2F%2Fmarketing-git-exp-performance-no3rd-rent-avail.vercel.app%2Finfo%2Fbigger-pockets-sponsorship

src/components/organisms/PitchCards/pitch-cards.stories.js
```diff
@@ -36,10 +36,10 @@ const sections = sectionData.map(({ title, description }) => ({
...
 
 const icons = [
-  { url: "/send-money.png", alt: "" },
-  { url: "/state-specific.png", alt: "" },
-  { url: "/user-card.png", alt: "" },
-  { url: "/property-damage.png", alt: "" },
+  <img src="/send-money.png" width="100%" alt="" />,
+  <img src="/state-specific.png" width="100%" alt="" />,
+  <img src="/user-card.png" width="100%" alt="" />,
+  <img src="/property-damage.png" width="100%" alt="" />,
```
pkrawc, line 42: Ah, I see here is the answer to my above question.

src/components/partials/AvailFooter/index.js
```diff
@@ -103,12 +104,12 @@ const AvailFooter = ({ animationPreset = "fadeIn" }) => {
...
                   href={urlResolver("https://www.avail.co/")}
                 >
-                  <Box
-                    as="img"
+                  <Image
                     src="/logo-glyph.svg"
-                    width="4rem"
-                    height="4rem"
+                    width={48}
+                    height={48}
```
pkrawc, line 110: Do REM values not work for the Image component?

src/utils/analytics/context.js
```diff
@@ -0,0 +1,22 @@
...
+  const scriptAppended = useRef(false)
+  return (
+    <AnalyticsContext.Provider
+      value={{
+        analytics,
+        setAnalytics,
+        analyticsPageParams,
+        setAnalyticsPageParams,
+        scriptAppended,
+      }}
```
pkrawc, line 17: This might be where we want to use `useMemo` to make sure that the analytics provider isn't rerendering children unnecessarily.

pkrawc (commented):
I'm not sure that allowing the Image to be passed in as an indiscriminate node to a bunch of organisms is the best idea.

In the organism something like below is more typesafe for me.

```jsx
// Pass the src, width, height, alt, and whatever else, but require the node to be an Image tag.
<Box as={Image} {...imageProps}  />
```

pages/info/[uid].js
```diff
@@ -96,21 +101,16 @@ const Page = ({ data, uid }) => {
...
+  const urlResolverParams = useMemo(() => {
+    return {
+      channel: data.query_channel,
+      utm_content: data.query_content,
+      signup_page: data.query_signup_page,
+      utm_campaign: data.query_campaign,
+      utm_source: data.query_source,
+      utm_medium: data.query_medium,
+    }
+  }, [data])
```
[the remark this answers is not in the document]
Comment (quaardvark) [reply pages/info/[uid].js:113]: It's not for props per se, it's for `useEffect` below that sets a state and depends on this object. Without `useMemo` it will create a render loop, because every render it would be a new object

src/components/partials/AvailFooter/index.js
quaardvark: Nope, has to be integer

quaardvark:
@pkrawc `next/image` relies on Next server to, well, serve it. You can't use it in non-Next context, i.e. storybook. So we can't really use the in organisms directly, which I personally don't mind at all, - we have the same pattern with RichText. 

Cloning the passed element is also kinda useless, - `style` is not supported for some reason. What works wonders though is `sx={{ "& img" : { ... } }}` on the parent.

src/utils/analytics/context.js
quaardvark: Maybe I'm dumb but I don't see any gains from memoing the value: https://codesandbox.io/s/quirky-browser-yf3bi?file=/src/App.js

I know people do this, but AFAIK it does nothing

quaardvark:
@pkrawc do you think this is mergeable?

src/utils/analytics/context.js
pkrawc: Cool. I often do it for contexts that have values that are want to change a lot, but for most scenarios, one does actually want children to re-render when the context changes.

pages/info/[uid].js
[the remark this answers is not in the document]
Comment (pkrawc) [reply pages/info/[uid].js:113]: Understood. Wouldn't setting these as a config of the hook itself be better then?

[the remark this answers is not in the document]
Comment (pkrawc) [reply pages/info/[uid].js:113]: Above, when you're calling the context make it a custom hook.

```js
const analytics = useAnalytics({ content: data.content, signup_page: data.signup_page, ...etc })
```
Which should replace the need for calling `useAnalyticsScript` too.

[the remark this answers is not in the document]
Comment (pkrawc) [reply pages/info/[uid].js:113]: Should run on every re-render (as intended by the useEffect hook).

pkrawc (commented):
A couple more questions before being ready to merge.

pages/info/[uid].js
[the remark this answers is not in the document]
Comment (quaardvark) [reply pages/info/[uid].js:113]: Why would analytics hook know anything about content and whatever else is it in Prismic? This is a serious abstraction leak. 

I see your general point though, I'll cook something to improve the API here

pkrawc (approved):
I'm fine with whatever you think is the best. I think I would abstract the script loading and analytics calling a bit. It doesn't seem so hyper-detailed that Prismic would have props loaded statically onto a page and then passed to an analytics call (whether segment of something else).

My initial thought was something like the below.

https://codesandbox.io/s/use-analytics-47kvg?file=/src/index.js

quaardvark:
@pkrawc I kinda see what you are getting at, but I see the script, the JWT decoding and the global variable (for Digioh) as a pure side effect, - don't want to smear it's details all over the app, thus I put it all in one hook. I don't think `useAnalytics` consumers should be concerned with what is loaded and how, it is purely a coincidence it has any kind of loading at all.

Also it has nothing to do with loading segment, but with analytics data from the main app. Segment is read from global var and stored in ref.

PR #100, opened 2021-05-25
Implementing Article List component

henriquecfreitas:
localhost_4000_info_henrique-test
localhost_4000_info_henrique-test(iPhone X)


pkrawc (changes requested):

quaardvark (changes requested):
I was not able to re-create the article list from the home page successfully (cc @pkrawc )

Screenshot_20210615_164954

1) Slice is missing description field. Skew?
2) Title should not have a green background. Title should be title and green background text should be an additional field eyebrow or tag
3) Images should have gray backdrop by default. Should also probably incorporate `next/image`
4) There are 2 fields on the slice that say "Background color" - not clear which is which

henriquecfreitas:
About the SS: I think you only need to set different color for the slice BG and the cards BG's

1. I'll add description and skew fields
2. I'll rename it to `tag`, we aren't handling `title` as a card field because each card can style it's title on it, or don't have title at all
3. Agreed
4. We have a `Background color` field and a `Background color for 'Read More' card` field. I personally think it is clear, considering the limitations we have to create a non repeatable section for the `Read More card`. I'm open for suggestions on that

@quaardvark 
> I was not able to re-create the article list from the home page successfully (cc @pkrawc )
> 
> Screenshot_20210615_164954
> 
> 1. Slice is missing description field. Skew?
> 2. Title should not have a green background. Title should be title and green background text should be an additional field eyebrow or tag
> 3. Images should have gray backdrop by default. Should also probably incorporate `next/image`
> 4. There are 2 fields on the slice that say "Background color" - not clear which is which


@pkrawc 
> * Description as in something to tell the Prismic user how to consume this? Or description as in text under the slice title?
> * That's a tag, not a title. Each article should probably have a title though.
> * Agreed
> * Agreed

quaardvark:
> About the SS: I think you only need to set different color for the slice BG and the cards BG's.
> And we have a bug on the button for large length labels, I'll work on that.
> 
> 1. I'll add description and skew fields
> 2. I'll rename it to `tag`, we aren't handling `title` as a card field because each card can style it's title on it, or don't have title at all
> 3. Agreed
> 4. We have a `Background color` field and a `Background color for 'Read More' card` field. I personally think it is clear, considering the limitations we have to create a non repeatable section for the `Read More card`. I'm open for suggestions on that
> 

1. :+1: 
2. We should have a title field, - we want a guided experience for editors. Styling can be done with extra labels, and / or, derived from card type. No title should be still possible with dedicated field
3. :+1: 
4. 
Screenshot_20210622_113149

henriquecfreitas:
> > About the SS: I think you only need to set different color for the slice BG and the cards BG's.
> > And we have a bug on the button for large length labels, I'll work on that.
> > 
> > 1. I'll add description and skew fields
> > 2. I'll rename it to `tag`, we aren't handling `title` as a card field because each card can style it's title on it, or don't have title at all
> > 3. Agreed
> > 4. We have a `Background color` field and a `Background color for 'Read More' card` field. I personally think it is clear, considering the limitations we have to create a non repeatable section for the `Read More card`. I'm open for suggestions on that
> 
> 1. 👍
> 2. We should have a title field, - we want a guided experience for editors. Styling can be done with extra labels, and / or, derived from card type. No title should be still possible with dedicated field
> 3. 👍
> 
> Screenshot_20210622_113149

We're seeing different labels in prismic...

Renaming `Background color` to `Card background color` its a good suggestion, I'll implement as well as the other labels suggestions.

Screen Shot 2021-06-22 at 8 22 04 AM

quaardvark:
@henriquecfreitas I'm looking at the Prismic type from this PR specifically, you can see it matches the names from my screenshot

src/components/organisms/ArticleList/article-card.js
```diff
@@ -19,14 +29,24 @@ function ArticleCard({ bg, link, image, title, content, action, ...props }) {
...
+            height="50%"
+            maxHeight="16rem"
+            {...image}
+            sx={{
+              "& *": {
+                height: "100%",
+              },
+            }}
+          >
+            <Image
```
quaardvark, line 44: We can't use `next/image` here, - it wouldn't work outside on Next.JS context. We should use `next/image` in Slice and `` in storybooks, just like in all other `next/image` usages in this repo

src/types/info.json
```diff
@@ -283,23 +321,23 @@
                 "type" : "Select",
                 "config" : {
                   "options" : [ "transparent", "ui_100", "ui_300", "ui_500", "ui_700", "ui_900", "blue_100", "blue_300", "blue_500", "blue_700", "blue_900" ],
-                  "default_value" : "transparent",
-                  "label" : "Background color"
+                  "default_value" : "ui_300",
+                  "label" : "Image background color"
```
quaardvark, line 325: Do we really need that? I think we agreed that it should be gray by default

quaardvark (changes requested):
FYI storybook for ArticleList is crashing

src/types/info.json
henriquecfreitas: It is gray by default now.
Giving the user the possibility to opt for another color instead of the default one, for the cases where gray isn't a good fit (eg: if you have the same gray as the bg color for the card)

quaardvark: Well, it that case shouldn't it be `ui_500`? I think `ui_300` is the default page background, so essentially it will blend with the page in this case?

quaardvark: Also please group link and image fields together (image first and link last)

src/components/organisms/ArticleList/article-card.js
```diff
@@ -0,0 +1,101 @@
...
+          </Box>
+        )}
+        {title &&
+          cloneElement(title, {
+            mb: "0.667rem",
+            sx: {
+              textAlign: "left",
+              ...title.props?.sx,
+            },
+          })}
```
quaardvark, line 76: Doesn't title have any default styling for font-size etc?

src/types/info.json
```diff
@@ -155,6 +156,193 @@
...
+              "title" : {
+                "type" : "StructuredText",
+                "config" : {
+                  "label" : "Article title",
+                  "placeholder" : "Title",
+                  "allowTargetBlank" : true,
+                  "multi" : "paragraph, heading3, heading4, heading5, heading6, strong, em, hyperlink, list-item, o-list-item, embed",
+                  "labels" : [ "hero", "headline", "title", "subtitle", "body__emphasis", "body", "small" ]
+                }
+              },
```
quaardvark, line 294: Should probably be single line headers? I don't see the point of paragraphs, lists and especially embeds

src/types/info.json
henriquecfreitas: Yeah, this was a mistake 😬
Already replaced to `ui_500`

src/components/organisms/ArticleList/article-link.js
```diff
@@ -0,0 +1,12 @@
+import React from "react"
+import NextLink from "next/link"
+
+function HandledLink({ link: Link, ...props }) {
+  return typeof Link === "function" ? (
+    <Link {...props} />
+  ) : (
+    <NextLink passHref href={Link.href} {...props} />
```
quaardvark, line 8: Abstraction leak, - we should avoid using `next` components in organisms

src/components/partials/SliceZone/components/ArticleListSlice/index.js
```diff
@@ -0,0 +1,107 @@
...
+      },
+      idx
+    ) => ({
+      uid: `${link.id}-${tag}-${idx}`,
+      bg: cardBg,
+      tag,
+      title: (
+        <Box>
+          <RichText render={cardTitle} />
+        </Box>
```
quaardvark, line 47: This should be done differently, - default font styling will be applied to the `<Box />` but the defaults in `RichText` for H1...H5 will override that. Please refactor in a way where RichText will be directly passed, w/o a `Box`

src/components/organisms/ArticleList/article-card.js
```diff
@@ -0,0 +1,103 @@
+import React, { cloneElement } from "react"
+import { Box } from "@rent_avail/layout"
+import Button from "components/elements/Button"
+import Card from "components/elements/Card"
+import { STYLING } from "config"
+import ArticleLink from "./article-link"
+
+function ArticleCard({
+  bg,
```
quaardvark, line 9: Should probably default to `ui_100`

src/components/organisms/ArticleList/article-list.stories.js
```diff
@@ -0,0 +1,221 @@
+import React from "react"
+import { Box } from "@rent_avail/layout"
+import { ArticleList } from "./index"
+
+export default { title: "Components/ArticleList" }
+
+export function Default() {
+  return (
+    <Box width="100%" py="4rem" background="#F9F9F9">
```
quaardvark, line 9: `F9F9F9` doesn't have enough contrast. `ui_300` is `#f3f3f3`. Should probably use `bg="ui_300"` though

src/components/partials/SliceZone/components/ArticleListSlice/index.js
```diff
@@ -0,0 +1,107 @@
...
+      <Box>
+        <RichText render={readMoreCardContent} />
+      </Box>
+    ),
+    action: {
+      type: "button",
+      label: readMoreCardLinkLabel,
+    },
+    link: (props) => <Link link={readMoreCardLink} {...props} />,
+  }
```
quaardvark, line 92: Should it still render when all the data is not filled in? Can we provide some sensible defaults in case we really want it to always be there? (cc: @pkrawc )

src/components/partials/SliceZone/components/ArticleListSlice/index.js
```diff
@@ -0,0 +1,107 @@
...
+        </Box>
+      ),
+      content: (
+        <Box>
+          <RichText render={content} />
+        </Box>
+      ),
+      action: linkType !== "none" && {
+        type: linkType,
+        label: linkLabel,
```
quaardvark, line 56: I'd prefer forming link as an element in the slice instead of this

src/components/organisms/ArticleList/index.js
```diff
@@ -0,0 +1,60 @@
...
+        maxWidth={containerWidth}
+        py="6rem"
+      >
+        {title &&
+          cloneElement(title, {
+            as: motion.aside,
+            ...animation?.item,
+            sx: {
+              ...title.props?.sx,
+              ...STYLING.title,
```
quaardvark, line 38: Default styling should be applied before `sx` in case it would need to be overridden by the caller. Same below with the description

src/components/organisms/ArticleList/article-list.stories.js
```diff
@@ -0,0 +1,221 @@
...
+              <Box>
+                <b>Suspendisse imperdiet auctor felis in hendrerit.</b>
+                <p>
+                  Donec sollicitudin dolor venenatis magna venenatis, vel
+                  viverra ligula rutrum. Interdum et malesuada fames ac ante
+                  ipsum primis in faucibus. Pellentesque egestas ex sit amet
+                  faucibus pretium.
+                </p>
+              </Box>
+            ),
```
quaardvark, line 189: We probably want to implement titles via separate API, otherwise this story doesn't reflect the slice

quaardvark (changes requested):

src/components/partials/SliceZone/components/ArticleListSlice/index.js
quaardvark: FYI @pkrawc I implemented a change where read more card is not rendered at all if there's no content

quaardvark:
@nshonda tagging you for review since you are a stakeholder for this item

quaardvark (approved):

nshonda (approved):

PR #105, opened 2021-06-03
Implementing How It Works carousel

henriquecfreitas:


src/components/molecules/Slider/index.js
```diff
@@ -4,9 +4,9 @@ import { ChevronLeft, ChevronRight } from "react-feather"
 
 import { Box } from "@rent_avail/layout"
 
+import ControlChevron from "components/elements/ControlChevron"
```
pkrawc, line 7: The filename should still be lowercase with a hyphen delineating spaces.

src/components/organisms/Testimonials/carousel/testimonials-carousel-item.js
```diff
@@ -10,12 +10,12 @@ function TestimonialsCarouselItem({
...
   ...props
 }) {
-  const [opacity, scale] = (() => {
+  const [opacity, scale] = useMemo(() => {
     if (level >= 3) return [0.9, 0.5625]
     if (level === 2) return [0.75, 0.625]
     if (level === 1) return [0.5, 0.75]
     return [0, 1]
-  })()
+  }, [level])
```
pkrawc, line 18: This is a great use of useMemo.

pkrawc (changes requested):
`src/components/organisms/HowItWorks/carousel/index.js`
Any reason the above isn't just a carousel named file? I think index files everywhere is a code smell.

`src/components/organisms/HowItWorks/slices/index.js`
I don't think we should be relating anything to slices to the organism.

src/components/molecules/Carousel/use-carousel.js
```diff
@@ -32,30 +32,33 @@ function useCarousel(items, {
...
 
-  const activeItem = items.length > activeIndex ? items[activeIndex] : {}
-  const visibleItems = [
-    visibleItem(-2),
-    visibleItem(-1),
-    visibleItem(0),
-    visibleItem(1),
-    visibleItem(2),
-  ]
+  const { item: activeItem } = useMemo(visibleItem, [visibleItem])
```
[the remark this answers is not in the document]
henriquecfreitas: `visibleItem` callback will be memorized, but I think that without wrapping its execution inside the `useMemo` it will be reexecuted on every render (but I might be wrong)

Taking a second look on the code now, I think that `activeItem` and `visibleItems` can be both defined on the same `useMemo`, having `visibleItem` declared as function inside the callback func, since it is not being exported

src/components/molecules/Slider/index.js
pkrawc: file and folder names should use the same convention

quaardvark:
@henriquecfreitas build is failing FYI

pkrawc:
Looks like this build is still failing @henriquecfreitas

pkrawc (changes requested):
Looks like the build is still failing, could we address that?

Mostly the changes I was asking for are code convention and cleanliness, so this is close to done IMO.

src/components/molecules/Carousel/use-carousel.js
```diff
@@ -32,30 +32,56 @@ function useCarousel(items, {
...
+  const { activeItem, visibleItems } = useMemo(
+    () =>
+      Array(visibleItemsLenght)
+        .fill(activeIndex - Math.floor(visibleItemsLenght / 2)) // -2 -2 -2 -2 -2
+        .reduce(
+          (returnObj, lowerVisibleIdx, fromLowerIdx) => {
+            const fromActiveIndex = lowerVisibleIdx + fromLowerIdx // -2 -1 0 1 2
+            const currentItem = visibleItem(fromActiveIndex)
+            returnObj.visibleItems.push(currentItem)
+            console.log(currentItem.level !== 0, currentItem)
```
quaardvark, line 61: Forgotten hacks

src/components/molecules/Carousel/use-carousel.js
```diff
@@ -32,30 +32,56 @@ function useCarousel(items, {
...
+          if (disableIntervalOnSelect) {
+            setCurrentInterval(0)
+          }
+        },
+      }
+    },
+    [getSafeIndex, activeIndex, setActiveIndex, setCurrentInterval]
+  )
+
+  const { activeItem, visibleItems } = useMemo(
```
quaardvark, line 52: I think `useMemo` is not the right approach here, since it will be recalculated on every click/transition anyway. I think we maybe want to put it in `useCallback` and pass `activeIndex` as argument instead of taking in from the scope. This way it will be very stable

src/components/molecules/Carousel/use-carousel.js
```diff
@@ -32,30 +32,56 @@ function useCarousel(items, {
...
-  const visibleItems = [
-    visibleItem(-2),
-    visibleItem(-1),
-    visibleItem(0),
-    visibleItem(1),
-    visibleItem(2),
-  ]
-
-  return [ activeItem, visibleItems ]
+  const visibleItem = useCallback(
```
quaardvark, line 35: Let's name functions a bit more appropriately, this looks like a prop/variable. `getVisibleItem` is probably clearer

quaardvark (changes requested):
I tried to check `Carousel` in Storybook but couldn't figure out what it supposed to do. Can you check if it needs changes per this PR? Or what is it supposed to be?

nshonda (approved):

nshonda:
@quaardvark we'll have some conflict when merging pages/main in here after the how it works skew changes

PR #107, opened 2021-06-07
Home page refactor

quaardvark:
With this PR i want give the pages project a new approach:

1) Structure data more or less exactly the way Prismic would structure it (sans additional functionality)
2) Use SliceZone instead of bespoke amalgamation of organisms (in most cases)
3) Make only discussed and agreed upon changes to organisms. If something was not discussed / unsure - bring it up to the group and we'll sort it out. In other words, - avoid unclear and not scoped changes. More likely than not, if something on legacy pages is not fully scoped for this project, - is because we have an alternative solution and/or decided to ditch it.

So far PR includes:
a) SkewBox refactor
b) Hero background image support
c) Homepage Hero reworked with the 1-3 in mind (as an example of project direction)


pkrawc:
For the skew specifically. I think we should move to use `clip-path` instead of an extra absolutely positioned element with a transform on it.

pkrawc:
So, looking through this. I'm not liking that non-pages are stored in the pages directory. Remember, every file in that directory automatically gets a route built for it. I'm actually surprised that pages like `home/data/hero` don't throw an error when nextjs is compiling the site.

pkrawc:
Ah, it's cause you changed the page extensions in the next config. Clever, any reason to go this route over keeping all non-route-based stuff out of pages and in the src folder?

quaardvark:
@pkrawc I don't remember already, I made a `static` folder in `src` for some reusable components which is probably more suitable. I agree, would be a tad cleaner to keep em out, especially since we are going to refactor all the throwaway components out.

nshonda:
@quaardvark home page is done, can you take a look?

quaardvark:
@nshonda looks good, but we probably want to also clean out the unused components in home folder + check image locations

nshonda:
@quaardvark removed unused components, I didn't see any wrong images path either

PR #113, opened 2021-06-10
Fix/pricing card updates

nshonda:
- left-aligned text fixes
- removed shadow from cards except the primary one
- added special offer field/tag

### DIFF
--- a/src/components/organisms/PlansPrices/index.js
+++ b/src/components/organisms/PlansPrices/index.js
@@ -14,11 +14,10 @@ const PlansGrid = styled(Grid)(
     variants: {
       vertical: {
         gridTemplateColumns: "1fr",
-        maxWidth: ["unset", "unset", "100%", "100%", "54rem"],
         mx: "auto",
       },
       horizontal: {
-        gridTemplateColumns: "repeat(auto-fit, minmax(24rem, 30rem))",
+        gridTemplateColumns: "repeat(auto-fit, 26rem)",
         justifyContent: "flex-start",
       },
     },
@@ -45,8 +44,11 @@ function PlansPrices({
         as={motion.aside}
         {...animation?.container}
         ref={intersectionView}
-        maxWidth={containerWidth}
-        py="6rem"
+        sx={{
+          py: "6rem",
+          maxWidth: ["unset", "unset", "100%", "100%", "54rem"],
+          boxSizing: "content-box",
+        }}
       >
         <Box textAlign="left">
           {(title || subtitle) && (
--- a/src/components/organisms/PlansPrices/index.js
+++ b/src/components/organisms/PlansPrices/index.js
@@ -25,6 +25,24 @@ const PlansGrid = styled(Grid)(
   })
 )
 
+const TextGrid = styled(Grid)(
+  variant({
+    prop: "direction",
+    variants: {
+      vertical: {
+        gridTemplateColumns: "1fr",
+        maxWidth: ["unset", "unset", "75%", "75%", "54rem"],
+        mx: "auto",
+      },
+      horizontal: {
+        gridTemplateColumns: "1fr",
+        maxWidth: ["unset", "unset", "75%", "75%", "61rem"],
+        mx: "auto",
+      },
+    },
+  })
+)
+
 function PlansPrices({
   bg,
   direction = "vertical",
@@ -45,38 +63,40 @@ function PlansPrices({
         as={motion.aside}
         {...animation?.container}
         ref={intersectionView}
-        maxWidth={["unset", "unset", "75%", "75%", "54rem"]}
+        maxWidth={containerWidth}
         py="6rem"
       >
         <Box textAlign="left">
-          {title && (
-            <motion.aside {...animation?.item}>
-              {cloneElement(title, {
-                mb: "1rem",
-                sx: {
-                  ...title.props?.sx,
-                  ...STYLING.headline,
-                },
-              })}
-            </motion.aside>
-          )}
-          {subtitle && (
-            <motion.aside {...animation?.item}>
-              {cloneElement(subtitle, {
-                mb: "1rem",
-                sx: {
-                  ...subtitle.props?.sx,
-                  ...STYLING.subtitle,
-                },
-              })}
-            </motion.aside>
-          )}
+          <TextGrid direction={direction}>
+            {title && (
+              <motion.aside {...animation?.item}>
+                {cloneElement(title, {
+                  mb: "1rem",
+                  sx: {
+                    ...title.props?.sx,
+                    ...STYLING.headline,
+                  },
+                })}
+              </motion.aside>
+            )}
+            {subtitle && (
+              <motion.aside {...animation?.item}>
+                {cloneElement(subtitle, {
+                  mb: "1rem",
+                  sx: {
+                    ...subtitle.props?.sx,
+                    ...STYLING.subtitle,
+                  },
+                })}
+              </motion.aside>
+            )}
+          </TextGrid>
           <PlansGrid direction={direction} gap="2rem" my="2rem">
-            {plans.map((plan, index) => (
+            {plans.map((plan) => (
               <PlanCard
                 direction={direction}
                 animationPreset={animationPreset}
-                shadow={index === 0 ? true : false}
+                shadow={plan.specialOffer}
                 key={plan.title}
                 {...plan}
               />
--- a/src/components/organisms/PlansPrices/index.js
+++ b/src/components/organisms/PlansPrices/index.js
@@ -73,8 +73,8 @@ function PlansPrices({
                 {cloneElement(title, {
                   mb: "1rem",
                   sx: {
-                    ...title.props?.sx,
                     ...STYLING.headline,
+                    ...title.props?.sx,
                   },
                 })}
               </motion.aside>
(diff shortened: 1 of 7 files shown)


src/components/organisms/PlansPrices/plan-card.js
```diff
@@ -56,15 +62,21 @@ export default function PlanCard({
...
+      sx={{ boxShadow : shadow ? 2 : 'none' }}
       color={color}
       border="none"
       ref={intersectionView}
     >
+      {specialOffer && <div style={{ position: 'absolute', top: '-1rem' }}>
+      <Tag mr="1rem" bg="gold_300" color="blue_500" sx={{ fontSize : '14px' }}>
+         <Award size={12} style={{ marginRight: '0.5rem' }}/> Special offer
+      </Tag>
+      </div>}
```
quaardvark, line 74: 1. We should use `<Box />` instead of `<div />` and `sx` prop instead of `style`.
2. "Special offer" text should be customizable.
3. Please use ESLint and Prettier, because this code is wrongly formatted.

src/components/organisms/PlansPrices/plans-prices.stories.js
```diff
@@ -43,16 +43,19 @@ const plans = [
     subtext: "Subtext",
     description: "Description",
     features: Features,
+    specialOffer: true,
```
quaardvark, line 46: Should probably be a rich text element

src/components/organisms/PlansPrices/index.js
```diff
@@ -72,10 +72,11 @@ function PlansPrices({
             </motion.aside>
           )}
           <PlansGrid direction={direction} gap="2rem" my="2rem">
-            {plans.map((plan) => (
+            {plans.map((plan, index) => (
               <PlanCard
                 direction={direction}
                 animationPreset={animationPreset}
+                shadow={index === 0 ? true : false}
```
quaardvark, line 79: Not sure if this is the right approach. I think the implication from design was that the "primary" card was the one with the "special offer" tag

quaardvark (changes requested):

src/components/organisms/PlansPrices/index.js
```diff
@@ -48,34 +66,37 @@ function PlansPrices({
...
+                })}
+              </motion.aside>
+            )}
+          </TextGrid>
           <PlansGrid direction={direction} gap="2rem" my="2rem">
             {plans.map((plan) => (
               <PlanCard
                 direction={direction}
                 animationPreset={animationPreset}
+                shadow={plan.specialOffer}
```
quaardvark, line 99: This will always be an object from Prismic, so always true

src/components/organisms/PlansPrices/index.js
```diff
@@ -25,6 +25,24 @@ const PlansGrid = styled(Grid)(
   })
 )
 
+const TextGrid = styled(Grid)(
+  variant({
+    prop: "direction",
+    variants: {
+      vertical: {
+        gridTemplateColumns: "1fr",
+        maxWidth: ["unset", "unset", "75%", "75%", "54rem"],
```
quaardvark, line 34: I still don't see where this comes from. 75% seems arbitrary to me

src/components/organisms/PlansPrices/plan-card.js
```diff
@@ -40,31 +42,46 @@ export default function PlanCard({
...
       border="none"
       ref={intersectionView}
     >
+      {specialOffer && (
+        <Box sx={{ position: "absolute", top: "-1rem" }}>
+          <Tag
+            mr="1rem"
+            bg="gold_300"
+            color="blue_500"
+            sx={{ fontSize: "14px" }}
```
quaardvark, line 73: Should be 1.1667rem

src/components/organisms/PlansPrices/index.js
```diff
@@ -48,34 +66,37 @@ function PlansPrices({
...
-          )}
+        <Box textAlign="left">
+          <TextGrid direction={direction}>
+            {title && (
+              <motion.aside {...animation?.item}>
+                {cloneElement(title, {
+                  mb: "1rem",
+                  sx: {
+                    ...title.props?.sx,
+                    ...STYLING.headline,
```
quaardvark, line 77: I think this is slipped from the original implementation. Pretty sure `sx` should be last here in order for labels to be able to override the default styling. Same with the `subtitle`

src/components/partials/SliceZone/components/PlansPricesSlice/index.js
```diff
@@ -27,10 +27,14 @@ const PlansPricesSlice = ({ slice }) => {
       buttonId,
       buttonIsPrimary,
       background: cardBackground,
+      specialOffer,
       ...props
     }) => ({
       ...props,
       background: cardBackground,
+      specialOffer: specialOffer.length > 0 && (
```
quaardvark, line 35: Can (very likely) also contain an empty string so an array length be 1, but no text. If you need to check use `specialOffer?.[0]?.text`

quaardvark (changes requested):
This has failed to compile, please keep an eye on that

src/components/organisms/PlansPrices/index.js
nshonda: is the same configurations from the PlansGrid, I just copied from there @quaardvark

src/components/organisms/PlansPrices/index.js
nshonda: this check if the object is set, so it throws true or false depending on it and adds the shadow css
Screen Shot 2021-06-14 at 12 43 47

src/components/partials/SliceZone/components/PlansPricesSlice/index.js
```diff
@@ -32,7 +32,7 @@ const PlansPricesSlice = ({ slice }) => {
     }) => ({
       ...props,
       background: cardBackground,
-      specialOffer: specialOffer.length > 0 && (
+      specialOffer: specialOffer && specialOffer?.[0]?.text && (
```
quaardvark, line 35: FYI, you don't need to do `specialOffer && specialOffer...` here, optional chaining takes care of that - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

src/components/organisms/PlansPrices/index.js
quaardvark: I think in this case we need to wrap both of them into the same container if possible. Not a good idea to keep two implementation of the same box, basically. Unless I'm missing something

src/components/organisms/PlansPrices/index.js
quaardvark: With the change in the slice - yes. I would coerce it to boolean for the clarity of intention

src/components/organisms/PlansPrices/index.js
nshonda: we need a separate box, otherwise, the text doesn't align correctly when we have horizontally stacked cards
below examples are using the plans grid
Screen Shot 2021-06-14 at 16 32 57

Screen Shot 2021-06-14 at 16 33 38

and here are using the custom text grid

Screen Shot 2021-06-14 at 16 36 03
Screen Shot 2021-06-14 at 16 35 23

quaardvark: @nshonda please implement a solution with a single box for sizing the container, refactor as much as needed. The flaw of current approach is very visible in a Storybook:

Screenshot_20210615_151749

Or on the tablet:
Screenshot_20210615_152551


Also, Figma has a max width of 54rem for both variants, so I really don't see where 75% or 61rem are coming from. There should be 2 sizes - 100% for screens where 2 horizontally stacked cards don't fit and 54rem for the rest.

If you have questions please reach out to me in Slack.

src/components/organisms/PlansPrices/index.js
quaardvark: `shadow={!!plan.specialOffer}`

src/components/organisms/PlansPrices/plan-card.js
quaardvark: This is not resolved, we should not use `px` anywhere

src/components/organisms/PlansPrices/plan-card.js
```diff
@@ -40,31 +42,46 @@ export default function PlanCard({
...
       bg={background}
+      sx={{ boxShadow: shadow ? 2 : "none" }}
       color={color}
       border="none"
       ref={intersectionView}
     >
+      {specialOffer && (
+        <Box sx={{ position: "absolute", top: "-1rem" }}>
+          <Tag
+            mr="1.1667rem"
```
quaardvark, line 70: This was 1rem, my comment about 1.1667rem was concerning the font size. I don't see a reason this element needs right margin.

src/components/organisms/PlansPrices/index.js
quaardvark: This is not resolved. Still major issues with layout
Screenshot_20210616_164005

Please implement a single sizing box, not 3 grids.

quaardvark:
@nshonda Still experiencing major issues with the layout

1) The text is not aligned with the pricing cards on vertical stacking
Screenshot_20210617_160531

2) Horizontally stacked is not centered on the page
Screenshot_20210617_160547

Let's connect tomorrow before the stand to try to resolve this properly

quaardvark (approved):

PR #117, opened 2021-06-14
Improvement: Another take on performance

quaardvark:
Pretty small gains, but

- Removed motion from header (nav bar) and footer
- Switched motion implementation to `LazyMotion`, upgraded to match `elements` version (2 motions in bundle otherwise)
- Refactored an inefficient ShowCase icon import (was importing all feather icons)
- Swapped async with defer for GTM and Segment scripts (async has priority in modern browsers if both are specified)


quaardvark:
@pkrawc @henriquecfreitas any takers?

pkrawc (approved):
Looks good to me. Didn't check all browsers to make sure `m` works as expected, but I don't see anything on the framer docs that would suggest otherwise.

_In case there was a polyfill not included in the slim version or something._

PR #118, opened 2021-06-15
Feat/extend cta

nshonda:
extend cta functionalities adding image and description fields


src/components/organisms/ButtonCTA/index.js
```diff
@@ -90,6 +99,11 @@ function ButtonCTA({
               {button}
             </Box>
           )}
+          {image && (
+            <Box as={motion.aside} {...animation?.item}>
+              {image}
+            </Box>
```
quaardvark, line 105: This needs more work. If I pass the image without explicit width / height it collapses to 0.

src/components/organisms/ButtonCTA/buttonCTA.stories.js
```diff
@@ -77,3 +79,19 @@ export function NoTitle() {
...
+        <img
+          height="160px"
+          width="160px"
+          src="https://avail-design-site.now.sh/images/building.svg"
+          alt=""
+        />
+      }
+    />
+  )
+}
```
quaardvark, line 97: This is not enough and looks really weird. Transparent image on white background. Please add more stories with image and use a different image

quaardvark (changes requested):

quaardvark (approved):

PR #120, opened 2021-06-16
Feat/add index option field

nshonda:
adding index meta option to prismic pages

### DIFF
--- a/pages/info/[uid].js
+++ b/pages/info/[uid].js
@@ -75,7 +75,7 @@ const Page = ({ data, uid }) => {
   const {
     meta_title: title,
     meta_description: description,
-    meta_indexable: indexable,
+    meta_indexable: indexable = true,
     meta_keywords: keywords,
     meta_image: image,
     nav_bar_type: navBarType,
--- a/pages/info/[uid].js
+++ b/pages/info/[uid].js
@@ -75,6 +75,7 @@ const Page = ({ data, uid }) => {
   const {
     meta_title: title,
     meta_description: description,
+    meta_indexable: indexable,
     meta_keywords: keywords,
     meta_image: image,
     nav_bar_type: navBarType,
@@ -129,7 +130,7 @@ const Page = ({ data, uid }) => {
       ...pageAnalyticsParams,
     }))
   }, [pageAnalyticsParams])
-
+  console.log(indexable)
   return (
     <React.Fragment>
       <BodyStyles bg={background} />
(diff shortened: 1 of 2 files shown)


pages/info/[uid].js
```diff
@@ -137,6 +138,8 @@ const Page = ({ data, uid }) => {
         <NextSeo
           title={title}
           description={description}
+          nofollow={!indexable}
+          noindex={!indexable}
```
Comment (henriquecfreitas) [inline pages/info/[uid].js:142]: Why don't create them as independent configuration booleans? In order to allow configurations like `index/nofollow`.
I'm not a SEO specialist so if what I'm saying makes no sense please ignore, but is it possible that we want to index a certain page and at the same time want its links not to be followed?

Comment (nshonda) [reply pages/info/[uid].js:142]: I suggested the same thing, but @quaardvark said a single toggle was enough

henriquecfreitas (approved):

src/types/info.json
```diff
@@ -1375,6 +1375,15 @@
...
     },
+    "meta_indexable": {
+      "type": "Boolean",
+      "config": {
+        "placeholder_false": "noindex/nofollow",
+        "placeholder_true": "index/follow",
+        "default_value": true,
+        "label": "Meta indexable"
+      }
+    },
```
quaardvark, line 1386: None of the pages currently have this value. What will happen to non-republished pages when this goes live? Will they still index or because of the missing value all pages will be not indexed?

pages/info/[uid].js
Comment (quaardvark) [reply pages/info/[uid].js:142]: @henriquecfreitas the request was only about indexing. But since Google can still follow links on noindex we add that too. I don't think having them separate is too useful, - better have a simpler editor experience. I think the only case where you'd want `noindex, follow` is the ugly sitemap you don't want users to see, but want Google to crawl.

src/types/info.json
nshonda: since we have the default value true they will remain indexed

quaardvark:
Could you test and confirm that? My understanding of Prismic was that it
does not add fields to older pages retroactively, so most production pages
would have this field as undefined until someone manually re-saves them and
re-publishes.

nshonda:
I tested on my prismic repository, i had an already published page, created the field, didn't do any changes to the page and the meta tag was set as index/follow

quaardvark:
Makes sense with the default added :)

Just be aware for the future features that Prismic doesn't add new fields
retroactively, including their default values.

PR #121, opened 2021-06-17
Changed the default background of the EmailCaptureInput

taylorsavegnago:
https://airtable.com/tblXKaz8x2gdERalp/viwuGoo8g6PEIozUb?blocks=bipHfSDXdVhEEkZrn

image
image


src/components/molecules/EmailCaptureInput/index.js
```diff
@@ -76,18 +76,19 @@ const EmailCaptureInput = ({
...
             },
             "&:focus-within": {
-              borderColor: isDark ? "ui_100" : "blue_500",
-              color: isDark ? "ui_100" : "blue_500",
+              borderColor: "ui_700",
+              color: "ui_700",
             },
             "&.filled:not(:focus-within):not(.error) .input__label-row": {
-              color: isDark ? "ui_300" : "ui_700",
+              color: "ui_700",
```
quaardvark, line 88: I think we should review those if we can remove them from here, because we need to keep the Input close to original. If needed - reach out to design team to confirm is the change was intentional

src/components/molecules/EmailCaptureInput/index.js
```diff
@@ -76,18 +76,19 @@ const EmailCaptureInput = ({
...
             },
             "& > input": {
+              backgroundColor: "ui_100",
```
quaardvark, line 91: Because we have a white background now we need to adjust buttons so that they are not too light on the white background

Screenshot_20210618_171516

In other words, when the input is on the `blue_500` background and the button is inside an input it should be `blue_500` primary, but `blue_100` when it's under the input.

quaardvark (changes requested):

src/components/molecules/EmailCaptureInput/index.js
```diff
@@ -76,18 +76,19 @@ const EmailCaptureInput = ({
         labelId={inputLabelId}
         sx={{
           "& > label": {
-            color: isDark ? "ui_300" : "ui_900",
+            color:  "ui_700",
```
quaardvark, line 79: Also there are 2 spaces, so please use ESLint and Prettier plugins that should fix that automatically

quaardvark (changes requested):
I think the last commit is not the right approach. I think all we need to do is to use `useTheme` and `useMediaQuery` hooks in order to find out if we are in the first breakpoint. Then, using this knowledge we can pass `background` prop to `Button` if we are within the first breakpoint (button rendered over `background`) or `ui_100` when we are not within the first breakpoint (button is rendered against the white input).

The last commit can be reverted in case this approach is implemented. I'd prefer a simple solution, since this issue will be touched by a larger scope concerning theming of the slices/organisms

quaardvark (approved):

PR #124, opened 2021-06-23
Possibility to include Tag and Image on Hero Rich Text

taylorsavegnago:
As we can see in example, now got the possibility to include image between the text and tags.

Captura de Tela 2021-06-23 às 10 09 25

To include image, it just click on the image icon.
Captura de Tela 2021-06-23 às 10 10 22

And to include tag, it just add a label tag.
Captura de Tela 2021-06-23 às 10 10 38

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "@rent_avail/input": "^0.4.5",
     "@rent_avail/layout": "^0.3.12",
     "@rent_avail/menu": "^0.2.8",
+    "@rent_avail/tag": "^0.1.20",
     "@rent_avail/typography": "^0.1.20",
     "@rent_avail/utils": "^0.5.0",
     "clsx": "^1.1.1",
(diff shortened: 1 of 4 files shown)


quaardvark (changes requested):
Thanks for the screenshots and trying to recreate the content we are targeting with this. This is awesome approach.

For the changes in `info.json` - can you please revert formatting? These strange spaces is how Prismic formats them and I need it to be the same when I merge the production type with diff tool. I think easiest way to keep it in check is to go back to Prismic and copy the JSON from their UI.

quaardvark:
@taylorsavegnago can you take a look on failing build? I think it's because `@rent_avail/tag` is missing in dependencies declaration

quaardvark:
@taylorsavegnago this PR need to be fixed

quaardvark (approved):

PR #125, opened 2021-06-23
adding cta with email capture

nshonda:
A new slice based on CTA slice that has an Email Capture instead of a button or image. Should probably convert the existing Email Capture slice to this new slice


quaardvark:
@nshonda, about the existing Slice - it's a great idea, I think we should do that. So instead of creating a new slice we need to extend and rename the old slice, keeping existing API and Slice key intact. Good thinking

src/types/info.json
```diff
@@ -296,18 +296,11 @@
...
+            "icon" : "call_to_action",
             "display" : "list",
             "non-repeat" : {
-              "hash" : {
-                "type" : "Text",
-                "config" : {
-                  "label" : "Jump link hash",
-                  "placeholder" : "A hash to link to this slice, e.g. - overview"
-                }
-              },
```
quaardvark, line 310: I think we should keep this?

quaardvark (changes requested):
This does not pass QA.

Pre-existing Slices  with these changes look like this:
Screenshot_20210629_190937

Should look like this:
Screenshot_20210629_191052

Steps to reproduce:
1) Connect you local env to `avail-marketing` repository
2) Compare https://info.avail.co/info/listings-100001 to http://localhost:4000/info/listings-100001

quaardvark (approved):

PR #126, opened 2021-06-24
Feat/how it works skew

nshonda:
As seen on Home:

Skew should wrap individual sections instead of whole slice
Add alternating background color feature

### DIFF
--- a/src/components/organisms/HowItWorks/index.js
+++ b/src/components/organisms/HowItWorks/index.js
@@ -22,7 +22,7 @@ function HowItWorks({
 }) {
   const [presets, intersectionView] = useInViewAnimation({ threshold: 0.05 })
   const animation = presets[animationPreset]
-  const isDark = analyzeColor(background)
+  const isDark = background ? analyzeColor(background) : false
   return (
     <Box
       as={motion.aside}
@@ -37,9 +37,11 @@ function HowItWorks({
           bg={background}
           as={motion.aside}
           {...animation?.container}
+          px="2rem"
           sx={{
             display: "flex",
             justifyContent: "center",
+            flexDirection: "column",
             maxWidth: containerWidth,
           }}
         >
--- a/src/components/organisms/HowItWorks/index.js
+++ b/src/components/organisms/HowItWorks/index.js
@@ -45,7 +45,7 @@ function HowItWorks({
               display: "grid",
               justifyContent: "center",
               flexDirection: "column",
-              maxWidth: containerWidth,
+              maxWidth: ["62rem", "62rem", "62rem", "80rem", "100rem"],
             }}
           >
             {eyebrow && (
--- a/src/components/organisms/HowItWorks/index.js
+++ b/src/components/organisms/HowItWorks/index.js
@@ -3,6 +3,7 @@ import { motion } from "framer-motion"
 import { useInViewAnimation } from "utils/animation"
 import { Box, Container, Grid, Col } from "@rent_avail/layout"
 import { Text } from "@rent_avail/typography"
+import SkewBox from "components/molecules/SkewBox"
 import Video from "components/elements/Video"
 import { STYLING } from "config"
 
@@ -12,6 +13,8 @@ function HowItWorks({
   sections = [],
   alternate = (idx) => idx % 2 !== 0,
   containerWidth,
+  alternateBackground = false,
+  background,
   animationPreset = "fadeIn",
   ...props
 }) {
@@ -25,7 +28,7 @@ function HowItWorks({
       {...props}
       ref={intersectionView}
     >
-      <Container {...(containerWidth && { maxWidth: containerWidth })}>
+      <Container sx={{ maxWidth: "none" }}>
         {eyebrow && (
           <Text
             as={motion.aside}
@@ -46,8 +49,11 @@ function HowItWorks({
         )}
         {sections.map(({ uid, ...section }, idx) => (
           <HowItWorksSection
+            containerWidth={containerWidth}
             key={uid || idx}
             {...section}
+            background={background === "transparent" ? null : background}
+            alternateBackground={alternateBackground ? idx % 2 === 0 : false}
             flip={alternate(idx)}
             animationPreset={animationPreset}
             mb={idx !== sections.length - 1 ? "6rem" : 0}
@@ -63,8 +69,11 @@ function HowItWorksSection({
   image = null,
   video,
   embed,
+  background,
+  alternateBackground,
   flip,
   animationPreset,
+  containerWidth,
   mb,
 }) {
   const copyColumn = ["span 12", flip ? "7 / span 6" : "1 / span 6"]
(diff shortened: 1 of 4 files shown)


src/components/organisms/HowItWorks/index.js
```diff
@@ -75,25 +84,75 @@ function HowItWorksSection({
     threshold: 0.25,
   })
   const animation = presets[animationPreset]
-
+  if (alternateBackground) {
```
quaardvark, line 87: Is it possible to do it without so much code duplication? What makes alternateBackground so different?

src/types/info.json
```diff
@@ -775,6 +775,15 @@
...
                   "label" : "First section media position"
                 }
+              },
+              "alternateBackground" : {
+                "type" : "Boolean",
+                "config" : {
+                  "placeholder_false" : "False",
+                  "placeholder_true" : "True",
+                  "default_value" : false,
+                  "label" : "Alternate background color? (will start with the color set on background color setting)"
```
quaardvark, line 785: Good label :+1:

src/components/organisms/HowItWorks/index.js
```diff
@@ -75,25 +84,75 @@ function HowItWorksSection({
...
+    return (
+      <SkewBox
+        color={background === "blue_500" ? "inherit" : "blue_100"}
+        skew="left"
```
quaardvark, line 91: Should probably expose it as param and add it to Prismic as well instead of hardcoding left

src/components/organisms/HowItWorks/index.js
```diff
@@ -75,25 +84,75 @@ function HowItWorksSection({
...
+        bg={background === "blue_500" ? "transparent" : "blue_500"}
```
quaardvark, line 92: Is `blue_500` the only option for background? If yes, we should add more.

src/components/organisms/HowItWorks/index.js
```diff
@@ -25,7 +28,7 @@ function HowItWorks({
       {...props}
       ref={intersectionView}
     >
-      <Container {...(containerWidth && { maxWidth: containerWidth })}>
+      <Container sx={{ maxWidth: "none" }}>
```
quaardvark, line 31: Should probably move it lower down the stack instead of doing this. I suggest to try `<Container as={Grid} ... />` or vice versa in each section and change this one to a box instead

src/components/organisms/HowItWorks/index.js
```diff
@@ -75,25 +84,75 @@ function HowItWorksSection({
...
   })
   const animation = presets[animationPreset]
-
+  if (alternateBackground) {
+    return (
+      <SkewBox
+        color={background === "blue_500" ? "inherit" : "blue_100"}
+        skew="left"
+        bg={background === "blue_500" ? "transparent" : "blue_500"}
+        as={motion.aside}
```
quaardvark, line 93: This is useless without `animation.{item|container}` props. I think `{...animation?.container}` should move from grid here

quaardvark (changes requested):

src/components/organisms/HowItWorks/index.js
nshonda: right now it's either transparent or blue_500, so we just alternate between those two, if we add more we'll probably need to add a second field when having alternating backgrounds I think

src/components/organisms/HowItWorks/index.js
nshonda: I replaced this for a box instead

src/components/organisms/HowItWorks/index.js
quaardvark: This is not resolved, I don't see this change implemented.

We should alternate between specified background and transparent.

quaardvark (changes requested):
Please implement customizable background options (all the default colors we currently use elewhere). I'm not commenting on code flaws there because I expect the to be weeded out during this refactor.

We are implementing it for the Homepage, which alternates between `blue_100` and `transparent`.

src/components/organisms/HowItWorks/index.js
```diff
@@ -1,59 +1,83 @@
...
-          <motion.aside {...animation?.item}>
-            {cloneElement(title, {
-              mb: "4rem",
-              sx: { ...STYLING.headline, ...title.props?.sx },
-            })}
-          </motion.aside>
-        )}
+      <Box>
+        <SkewBox
+          color={isDark ? "blue_100" : "inherit"}
```
quaardvark, line 35: We probably want `ui_100` here until theming arrives

src/components/organisms/HowItWorks/index.js
```diff
@@ -1,59 +1,83 @@
...
+            </motion.aside>
+          )}
+        </SkewBox>
         {sections.map(({ uid, ...section }, idx) => (
           <HowItWorksSection
+            containerWidth={containerWidth}
             key={uid || idx}
             {...section}
+            skew={skew}
+            background={background === "transparent" ? null : background}
```
quaardvark, line 73: `background: transparent` is a valid value, should be no need for `null` here

src/components/organisms/HowItWorks/index.js
```diff
@@ -1,59 +1,83 @@
...
-        {title && (
-          <motion.aside {...animation?.item}>
-            {cloneElement(title, {
-              mb: "4rem",
-              sx: { ...STYLING.headline, ...title.props?.sx },
-            })}
-          </motion.aside>
-        )}
+      <Box>
+        <SkewBox
```
quaardvark, line 34: I'm not sure we need SkewBox here, since the children are also wrapped in it. What does this achieve with `skew=none`

quaardvark (changes requested):
As per previously discussed guidelines, please fix the storybook. Currently it does not reflect what we are trying to achieve.

1) Add a story with alternating backgrounds
2) Fix the "With Flip" story with strange colors
3) Fix the container width issue so that 

Additionally, please fix the regression on the page https://info.avail.co/info/product-referral-program-login (http://localhost:4000/info/product-referral-program-login)

image

quaardvark (changes requested):
The header is still not in the right place
Screenshot_20210706_172324

Should align with the rest of the components on the page

src/components/organisms/HowItWorks/index.js
```diff
@@ -1,59 +1,87 @@
...
+          sx={{
+            width: "100%",
+          }}
+        >
+          <Box
+            sx={{
+              display: "grid",
+              justifyContent: "center",
+              flexDirection: "column",
+              maxWidth: ["62rem", "62rem", "62rem", "80rem", "100rem"],
```
quaardvark, line 48: This breaks the organism abstraction. We should use the `containerWidth` value passed from the slice here

quaardvark (changes requested):
Still doesn't align on 1920px screen

image

src/components/organisms/HowItWorks/index.js
quaardvark: @nshonda 

There's some serious issues with the header layout. It looks like it's not centered on the page, but it centers it's content intsead

image

I'd suggest to use `Container` with `containerWidth` - should align fine

quaardvark (approved):

PR #127, opened 2021-06-25
Pattern padding on top and bottom between sections

taylorsavegnago:
As we can see in the example, it was added a padding pattern on top and bottom of the sections. This is for all of components.

Captura de Tela 2021-06-25 às 10 07 16

### DIFF
--- a/src/components/organisms/Hero/index.js
+++ b/src/components/organisms/Hero/index.js
@@ -65,7 +65,6 @@ function Hero({
         ref={intersectionView}
         as={Grid}
         minHeight={stretch && "calc(90vh - 14rem)"}
-        py={skew === "none" && !stretch && "2rem"}
         alignItems="center"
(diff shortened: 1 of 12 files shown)


quaardvark:
@pkrawc @taylorsavegnago  can't ping Vinnie here, but there's an issue with this when Skew is involved. Skew may clip the content with static padding
Screenshot_20210625_164956

It clips at the bottom and at the same time the top is misproportionally large (because it transforms from the top)

Not sure if this a design problem yet, just throwing it in. We might to put a pin in this before we rethink/refactor SkewBox

pkrawc:
Yeah. If I understand the problem here it's why I'm suggesting using `clip-path` on SkewBox instead of another node that acts as the background.

t0nylombardi (approved):
This looks good to go. has this been staged for QA?

PR #128, opened 2021-07-01
Feat/landlords page

taylorsavegnago:
Following the example, I created the data and the page.

But the page has some issues that I believe are about the branch.

And I need the MailTo for the call to action

Hero
image


quaardvark (commented):
@taylorsavegnago Great job! We should probably clean up previously used components + images and this will be good to merge

PR #129, opened 2021-07-02
Created tenants page

taylorsavegnago:
Following the example, I created the data and the page.

And I need the MailTo for the call to action


quaardvark (changes requested):
Same as with landlords features page - we need to cleanup unused files and images

PR #133, opened 2021-07-08
pricing page refactor

nshonda:
created a new  simple text content slice, that receives a simple title and rich text content, I refactored the pricing page following the marketing pages spec from figma


quaardvark (commented):
Looks good, the only thing is that level 3 headers and some text coloring doesn't match the design spec. Could you double check please?

PR #137, opened 2021-07-19
Create Standard NavBar Organism

nshonda:
- Adding new Prismic types (Navbar and NavbarMenu)
- Using Avail/RDC navbar across all pages
- adding support to 1st level navigation
- adding tag support to child links
- copying existing menu to static navbar


nshonda:
@quaardvark leaving this as a draft PR for now

quaardvark:
@nshonda can you check why the build is failing? You can try locally with `yarn build`. It seems to be happening in `[uid].page.js` - I don't think this PR should contain any changes for this page

quaardvark (commented):
I'm not sure if I fully understand the flow of work here, but I though we were going to add a new NavBar organism instead of modifying existing NavBars that are already used on paid search pages.

nshonda:
@quaardvark ah I see, okay, there's no problem I can convert my changes into a new organism then and revert the existing NavBars to the state they were on pages/main branch

quaardvark (approved):

PR #138, opened 2021-07-22
Feat/custom columns sizes

nshonda:
- Custom column sizes for Hero and Hero with email capture
- Custom column sizes for How it Works
-  Improved Stories to visualize custom columns


src/components/organisms/HowItWorks/index.js
```diff
@@ -83,14 +86,55 @@ function HowItWorksSection({
...
+      case 8:
+        imageValue = '5 / span 8'
+        break;
+      case 4:
+          imageValue = '9 / span 4'
+          break;
+    }
+  } 
+  const copyColumn = ["span 12", copyValue]
+  const imageColumn = ["span 12", imageValue]
```
quaardvark, line 137: There are few things I think we can optimize here:

1) Move it as a function outside the component - this way the code will not be recreated every render. Use `return` instead of `break`
2) Instead of switching each span separately - switch on the whole string
3) Instead of repeating the switch statements for `flip` - do a ternary conditional assignment

```
function getColumnSpans(sizes) {
    switch (sizes) {
        case "8,4":
            return flip ? ["5 / span 8", "auto / span 4"] : ["auto / span 8", "9 / span 4"]
        # ...and so on
    }
}

# then in component

const [copySpan, imageSpan] = getColumnSpans(columnSizes)
```

src/components/partials/SliceZone/components/HeroSlice/index.js
```diff
@@ -28,9 +28,10 @@ const HeroSlice = ({ slice, sliceIndex }) => {
       secondaryButtonLink,
       secondaryButtonId,
       outro,
+      columnSize,
     },
   } = slice
-
+  console.log(columnSize)
```
quaardvark, line 34: Forgotten console

src/components/organisms/Hero/hero.stories.js
```diff
@@ -6,245 +6,169 @@ import { Box } from "@rent_avail/layout"
...
+    },
+    image: {table : { disable: true }},
+    imagePosition: {control: {type: 'select', options: ['left', 'right'] }},
+    animationPreset: {table : { disable: true }},
+    columnSize: {
+      control: { type: 'select', options: ["[8,4]",
+      "[6,6]",
+      "[4,8]"]}
+    },
+  }
```
quaardvark, line 38: Nice

quaardvark (changes requested):

quaardvark (approved):

PR #139, opened 2021-07-23
Created the Whats In A organism

taylorsavegnago:
Result: 
image


taylorsavegnago:
The effect developed is not 100% equals to the layout because the keyframes existents doesn't match, for this reason I needed to do some adjustments

src/components/organisms/WhatsInA/whats-in-a.stories.js
```diff
@@ -0,0 +1,92 @@
...
+    number: "5",
+    left: "30%",
+    top: "72%",
+  },
+  { 
+    number: "6",
+    left: "65%",
+    top: "90%",
+  },
+]
```
quaardvark, line 72: I think we should combine these arrays into one, something like 
```
{ 
    number: "6",
    location : {
        left: "65%",
        top: "90%",
    },
    title: "Disclosures in lease",
    text: "Besides attaching require disclosure pamphlets, the lease includes all required disclosure language directly within the lease.",
},
```

src/components/organisms/WhatsInA/index.js
```diff
@@ -0,0 +1,142 @@
...
+const Span = styled.div`
+  border-radius: 50%;
+  width: 27px; 
+  height: 27px;
+  padding: 1px 9px;
+  content: counter(step-counter); 
+  position: absolute;
+  font-size: 0.8em;
+  font-weight: 600; 
+  line-height: 27px;
```
quaardvark, line 135: We should try to avoid duplicating properties, - ideally these should be derived from the same base styled component

src/components/organisms/WhatsInA/index.js
```diff
@@ -0,0 +1,142 @@
...
+                    sx={{
+                      borderRadius: '50%', 
+                      width: '27px', 
+                      height: '27px',
+                      padding: '1px 9px',
+                      content: 'counter(step-counter)', 
+                      position: 'absolute', 
+                      fontSize: '0.8em', 
+                      fontWeight: '600', 
+                      lineHeight: '27px'
```
quaardvark, line 95: All `px` and `em` sizes should be converted to `rem`. Probably should round them up for convenience as well

src/components/organisms/WhatsInA/index.js
```diff
@@ -0,0 +1,142 @@
...
+  position: absolute;
+  font-size: 0.8em;
+  font-weight: 600; 
+  line-height: 27px;
+  color: #fff;
+  text-align: center;
+  background-color: #0f3e6f;
+  animation-name: ${pulseit};
+  animation-duration: 1.5s;
+  animation-iteration-count: infinite;
```
quaardvark, line 141: Can we do this with `sx` somehow?

src/components/organisms/WhatsInA/whats-in-a.stories.js
```diff
@@ -0,0 +1,92 @@
...
+
+export function Default() {
+  return (
+    <WhatsInA
+      title={<Box as="h1" color="blue_500">What's In A Title</Box>}
+      description={<Text as="h5" color="blue_700">What's In A Description</Text>}
```
quaardvark, line 78: Top :+1:

quaardvark (changes requested):
Styling looks broken on tablet
Screenshot_20210726_173324

and on the mobile
Screenshot_20210726_174944

Code formatting also doesn't look good, - please use ESLint and Prettier, - they report 77 issues
Screenshot_20210726_175129

quaardvark (approved):

PR #147, opened 2021-08-09
Created the Tenant Screening page

taylorsavegnago:
I just have a doubt, the Plans and Prices is not aligned with the other components? I got the actual component and it seems it hasn't the same width comparing to the other components, neither in prismic nor in storybook


nshonda:
@quaardvark @taylorsavegnago  I think we'll encounter a few similar design feedbacks regarding specifics components (like the Plan and Prices one) during QA for all the pages, do you think it's better to just check for content and syntax now and address these issues later?

nshonda (approved):

PR #161, opened 2021-08-24
Created the offset on Hero images

taylorsavegnago:
New prop on Hero images:

offset: {
  top: "0",
  right: "0",
  left: "0",
  bottom: "0"
},

It means where you want the image to be located in the Hero

image

image


quaardvark (changes requested):
I don't think it should be defined as offset, because it would be hard to use when implemented in Prismic. I think the Prismic data should be just `imagePosition: "offset-bottom"` and just make it stick to the bottom.

quaardvark (changes requested):
There's something weird is happening on different screen sizes

Screenshot_20210830_151159

taylorsavegnago:
> There's something weird is happening on different screen sizes
> 
> Screenshot_20210830_151159

I fixed the height of Hero according to the old page (https://www.avail.co/landlords/features). Used 56rem.

quaardvark (changes requested):
Please, do not change the hero height

quaardvark:
@nshonda @taylorsavegnago FYI this was implemented properly now, but it looks strange on big screens, so I asked Vinnie to review it

PR #165, opened 2021-09-02
Fixed the redirect in email capture input

taylorsavegnago:
The error was: Cannot read property 'uuid' of undefined

So I noticed that was missing the "pageUid" on each page.

I set it with "setAnalytics" and it worked.


quaardvark:
Closing in favor of #174

PR #176, opened 2021-09-14
fixing skipped heading levels on rent analysis page

nshonda:
https://airtable.com/appsEN1RgrUk3oyon/tbluJ0kJjMnu1rtsx/viwEjc1BEgM0pj72s/rec35p6TMX5CENGpf?blocks=hide


pages/landlords/rent-analysis/data/how-it-works.js
```diff
@@ -29,7 +29,7 @@ export default {
       },
       title: [
         {
-          type: "heading3",
+          type: "paragraph",
```
quaardvark, line 32: I think this should stay H3? Also this text is not legible

pages/landlords/rent-analysis/data/second-how-it-works.js
```diff
@@ -28,7 +28,7 @@ export default {
       },
       title: [
         {
-          type: "heading3",
+          type: "paragraph",
```
quaardvark, line 31: Same with this file, aren't they are supposed to be H3?

pages/landlords/rent-analysis/data/third-how-it-works.js
```diff
@@ -49,7 +49,7 @@ export default {
       },
       title: [
         {
-          type: "heading3",
+          type: "paragraph",
```
quaardvark, line 52: Same

pages/landlords/rent-analysis/data/how-it-works.js
nshonda: since the previous heading next to it is a h1 it needs to be either a h2 or a paragraph, https://www.w3.org/WAI/tutorials/page-structure/headings/

quaardvark: Can you verify that with marketing? I think a better solution would be to add an H2 title to the slice

quaardvark:
@nshonda sorry, forgot to close thos. This will be fixed in #196

PR #178, opened 2021-09-14
adding benefits variants

nshonda:
Adding a variant to pitch cards so content gets centered
related issue:
- https://airtable.com/appsEN1RgrUk3oyon/tbluJ0kJjMnu1rtsx/viwEjc1BEgM0pj72s/recQJIn1jnDLSZR3I?blocks=hide


quaardvark (changes requested):
I think we should add a select here with a `variant` key instead of a bool. We should also treat the other "button" variant the same way, so PitchCard api would look like `<PitchCard variant={"button" | "benefits"} />`

quaardvark (changes requested):
`variant` should a slice property, not each invidual pitch card

src/components/partials/SliceZone/components/PitchCardsSlice/index.js
```diff
@@ -53,7 +55,7 @@ const PitchCardsSlice = ({ slice }) => {
       key: title?.[0]?.text || idx,
       link: linkText ? (
         <Link link={linkUrl}>
-          {linkAsButton ? (
+          {variant === "button" ? (
```
quaardvark, line 58: This should also work with the old API, but maybe I missed where it's stubbed

nshonda:
@quaardvark do you think we'll have issues with the old API if we make `variant` be a slice property since `linkAsButton` was a card property?

quaardvark:
@nshonda All I ask is to hoist variant to top and implement this `const isButtonVariant = variant === "button" || linkAsButton`. We'll sort the API thing later, there's probably only a few pages that use it

quaardvark:
Closing in favor of #196

PR #197, opened 2021-09-22
Exported images from figma and substituted

taylorsavegnago:
Exported images from figma and substituted the image and the size according to the design


quaardvark:
@taylorsavegnago there's something wrong with the image on home hero. These images have weird positioning. I'll sort it out with winnie, putting this PR on hold

PR #203, opened 2021-09-28
Adding CI/CD pipeline

fboaventura:
Adding the CI/CD pipeline to build the Docker image, publish it to ECR and update the ECS cluster task with the new image.


quaardvark (commented):
We would need different env variables for prod and staging, including secrets. Is this achievable with this setup?

PR #204, opened 2021-09-30
Updating Fastpay copy content

nshonda:
Related issue: https://github.com/rentalutions/marketing/compare/fix-fastpay-copy?expand=1


pages/landlords/fastpay/data/pitch-cards.js
```diff
@@ -5,6 +5,7 @@ export default {
     hash: null,
     title: [],
     titleImage: {},
+    variant: "centered-benefits",
```
quaardvark, line 8: This conflicts with the design spec, please revert

quaardvark (changes requested):

PR #211, opened 2021-10-05
 Reenable animation of Hero on client-side transitions

quaardvark:


src/components/organisms/Hero/index.js
```diff
@@ -175,7 +175,7 @@ function Hero({
           >
             <Box
               as={motion.div}
-              {...animation?.item}
+              {...(!hasOffsetImage && animation?.item)}
```
quaardvark, line 178: Had to switch off the animation on offset images, because it conflicts with all the positioning hacking. Might fix later

PR #218, opened 2021-10-08
Add homepage redirect logic for logged in users and info pages

quaardvark:
Rentalutions homepage redirects logged in users based on their role. Since the new homepage is a static page and doesn't really have a way to analyze `httpOnly` cookies w/o basically turning off all the static page optimizations in NextJS, for logged-in users I added a native NextJS redirect that matches the presence of `_rentalutions_session` cookie and redirects user to an API route that will resolve a user (via forwarding cookies in a server-side request) and redirect based on the level. If it will not be able to reach a user, - we will unset the session cookie and redirect back to home.

This solution is somewhat fragile, because of the cookie forwarding (future security improvements may disallow this route) and messing with a session cookie at all.


next.config.js
```diff
@@ -16,6 +16,29 @@ const pageExtentions = (() => {
   }
 })()
 
+const PAGES_REWRITES = [
+  {
+    source: "/pages-misc/home-fallback",
+    destination: "/",
+  },
+]
```
quaardvark, line 24: Adds a fallback version of homepage for local development, in case rentalutions app is not running locally

next.config.js
```diff
@@ -16,6 +16,29 @@ const pageExtentions = (() => {
...
+
+const INFO_REWRITES = [
+  {
+    source: "/",
+    destination: "/index-info",
+  },
+]
```
quaardvark, line 31: Rewrite of an index page for `info.avail.co`

pages/misc/resolve-user.page.js
```diff
@@ -0,0 +1,53 @@
+import React, { useEffect } from "react"
+import { Box } from "@rent_avail/layout"
+import LoadingCircle from "components/elements/LoadingCircle"
+
+const FALLBACK_PATH =
+  process.env.NODE_ENV === "development"
+    ? "/pages-misc/home-fallback"
+    : "/app/landlords/dashboard"
```
quaardvark, line 8: rentalutions redirects to `/app/landlords/dashboard` as a fallback, so we are recreating this here

pages/api/resolve-user.api.js
```diff
@@ -0,0 +1,34 @@
...
+    headers: { cookie: req.headers.cookie },
+  })
+
+  function handleFallback() {
+    res.setHeader(
+      "Set-Cookie",
+      "_rentalutions_session=nil; path=/; expires=Thu, 01 Jan 1970 00:00:00 GMT"
+    )
+    res.redirect("/")
+  }
```
quaardvark, line 15: Since the Next redirect will loop over the `/` otherwise we need to unset the session cookie. On logout it is replaced with a session cookie not associated with a user, so I can't think of a better way of handling this

src/components/elements/LoadingCircle/index.js
```diff
@@ -0,0 +1,54 @@
+import React from "react"
```
quaardvark, line 1: This is not used and related to a previous version of solution, but keeping it, since it's eventually find it's use

PR #219, opened 2021-10-08
adding dynamic content

nshonda:
Allow us to have dynamic content tags inside rich text


nshonda:
@quaardvark please review, I tried to meet all the adjustments requested, if there's anything missing let's schedule a pair programming

pages/home/data/articlelist.js
```diff
@@ -80,7 +80,7 @@ export default {
     description: [
       {
         type: "paragraph",
-        text: "Join the 197,548+ landlords who use Avail to responsibly manage their properties.",
+        text: "Join the {LANDLORD_COUNT}+ landlords who use Avail to {LANDLORD_COUNT} responsibly manage their properties.",
```
quaardvark, line 83: I don't think count should be twice here

src/config.js
```diff
@@ -74,3 +74,7 @@ export const EMBED_WHITELIST = [
   "https://www.youtube.com/",
   "https://wistia.com",
 ]
+
+export const DYNAMIC_CONTENT = {
+  LANDLORD_COUNT: "281,056",
```
quaardvark, line 79: This should be a transforming function, not a primitive, IMO

PR #222, opened 2021-10-14
Add homepage redirect logic for logged in users and info pages, rework analytics traits

quaardvark:
This PR includes

1) Homepage redirects replication. This relies on `__AUT` (Avail user token) JWT cookie that Rails app would set. Since NextJS homepage is static and we don't want to turn it dynamic by using `getServerSideProps` we also need a static process to conduct a redirect and one of the ways to do that is a static NextJS router config that can redirect visitors based on the cookie presence. We redirect visitors with `__AUT` cookie to `/api/resolve-user` which will decode and verify the JWT containing all the information needed for redirect and perform a redirection.

2) Changes to traits processing for analytics, - this app now almost completely relies on Rails app for storage and processing of analytics traits. This was achieved by extending the  `__AVT` JWT cookie to pass all the traits available to Rails app to the marketing app. This makes the Rails app the single source of truth for analytics trains, minimizes the risk that the traits will become out of sync between apps and reduces the possibility of one app overwriting user traits previously submitted by other and vice versa.

Rails app will also use UUID cookie for fetching the user traits from DB even if user is not logged in, furthermore making this a decent solution.


next.config.js
```diff
@@ -16,6 +16,22 @@ const pageExtentions = (() => {
   }
 })()
 
+const INFO_REWRITES = [
+  {
+    source: "/",
+    destination: "/index-info",
+  },
+]
```
quaardvark, line 24: Rewrite for homepage index on `info.avail.co`. This is needed because info pages index needs a different file name, other than `index` to avoid collision with core pages index (homepage) when running in PAGES_TYPE=all mode (development).

next.config.js
```diff
@@ -16,6 +16,22 @@ const pageExtentions = (() => {
...
+
+const PAGES_REDIRECTS = [
+  {
+    source: "/",
+    has: [{ type: "cookie", key: "__AUT" }],
+    destination: "/pages-api/resolve-user",
+    permanent: false,
+  },
+]
```
quaardvark, line 33: Setup for redirect to `/api/resolve-user` based on the presence of `__AUT` cookie.

pages/api/resolve-user.api.js
```diff
@@ -0,0 +1,34 @@
...
+      return redirectUrl ? res.redirect(redirectUrl) : handleFallback()
+    } catch {
+      return handleFallback()
+    }
+  } else {
+    return handleFallback()
+  }
+}
+
+export default resolveUser
```
quaardvark, line 34: An API for user redirection based on `__AUT` payload

pages/api/resolve-user.api.js
```diff
@@ -0,0 +1,34 @@
...
+const AVAIL_ECDSA_PUBLIC_KEY = Buffer.from(
+  process.env.AVAIL_ECDSA_PUBLIC_KEY,
+  "base64"
+).toString()
+
+async function resolveUser(req, res) {
+  function handleFallback() {
+    res.setHeader(
+      "Set-Cookie",
+      "__AUT=; path=/; expires=Thu, 01 Jan 1970 00:00:00 GMT"
```
quaardvark, line 13: We need to unset the `__AUT` cookie if we can't redirect properly to avoid circular redirects

pages/index-info.info.js
```diff
@@ -0,0 +1,32 @@
...
+      },
+    }
+  }
+  return {
+    redirect: {
+      destination: "/info/listings",
+      permanent: false,
+    },
+  }
+}
```
quaardvark, line 32: The `info.avail.co` index page. Just adding it back, since it was overwritten by the core pages homepage.

src/components/elements/LoadingCircle/index.js
```diff
@@ -0,0 +1,54 @@
...
+        r={radius}
+        stroke="currentColor"
+        strokeWidth="4"
+        strokeLinecap="round"
+      />
+    </Svg>
+  )
+}
+
+export default LoadingCircle
```
quaardvark, line 54: Just an animated spinny circle I added in one of my previous attempts to solve the redirects problem. Not removing, since will probably be useful

src/components/partials/PageNavBar/index.js
```diff
@@ -43,6 +83,25 @@ function PageNavBar({ document, ...props }) {
...
+
+  useEffect(() => {
+    if (window.document.cookie.includes("__AUT")) {
+      setMyAccountUrl("/") // Reduces layout shift
+      resolveMyAccountUrl()
+    } else {
+      setShowSignupButtons(true)
+    }
+  }, [resolveMyAccountUrl])
+
```
quaardvark, line 104: A bit of hacky way to quickly implement `My Account` button for logged-in users. Keys off of presence of `__AUT` cookie as well, decodes the JWT and sets up a button with a right link (reuses the same logic as redirects)

src/utils/avail.js
```diff
@@ -0,0 +1,32 @@
...
+        ? "/app/landlords/my-first-unit"
+        : "/app/landlords/dashboard"
+    case "tenant":
+      return "/app/tenants/dashboard"
+    case "admin":
+      return "/admin"
+    default:
+      return null
+  }
+}
```
quaardvark, line 32: Redirect logic for homepage and `My Account` button

